PR #1, opened 2024-05-21
Get the Project Setup

coreyja:
Doing this as a PR so I can comment it up and explain stuff!

We are setting up the project as a Cargo Workspace with two crates at the moment:
- db
- cli

`db` connects to the database and contains the DB migrations
The `cli` crate is a binary crate and uses the `db` crate to acquire a pool

Requirements:
- Postgresql Installed
- DATABASE_URL set

### Postgres

If you are on OSX you can install postgres with `brew install postgresql` and if you are on another platform and need a hand installing postgres let me know and happy to help!

And then you will need to set the `DATABASE_URL` to something like this `DATABASE_URL="postgres://localhost:5432/knowledge"`
I use `direnv` so that I can create a `.envrc` file in the repo root and then when I'm in those directories I will get the DATABASE_URL set for me
https://direnv.net/

But there are lots of env var managers, and if you have a preference thats totally fine!
An even lower tech, non project specific answer could be to but that env var in your `.bash_profile` (or equivilent) but I like being able to have env vars per project!

### SQLx

We are going to use `sqlx` https://github.com/launchbadge/sqlx to integrate with Postgres! This is a cool library that actually does Rust type checking over the sql queries too.

Hopefully their docs are enough to get you started a bit, but PLEASE ask questions as you run into them lol!

### Migrations

`migrations` are how we change the Database 'schema'. Basically its how we add new tables, and new columns to our databse!

Here are the SQLX cli docs on how to create a run them: https://github.com/launchbadge/sqlx/blob/main/sqlx-cli/README.md#create-and-run-migrations

They are _basically_ little SQL scripts that get run when the server boots, IF they haven't run yet. SQLx tracks which ones have run for you and only runs new ones! This helps when you have multiple people working on the same DB, and makes sure everyone can have an up to date schema for their DB!
You'll definitely be making some migrations, so good to get familiar with the concepts a bit! I think even your first card will involve a migration to create a Users table!

I've created a `db/migrations` directory which we can use to store the migrations!


.github/workflows/rust_tests.yml
```diff
@@ -0,0 +1,54 @@
+name: Rust Tests
```
coreyja, line 1: Github Actions for CI Testing!

We are running:
- `cargo test` to run all the Rust Tests that we will write
- `cargo clippy` which is the Rust linter
- `cargo deny check bans` which will look for Crate (Rust packages) that I've explicitly banned. This is really just a ban on `openssl` right now, since that a `C` library and makes deploying more annoying than it needs to be. So we try to always use `rustls` which is a full Rust implementation. And this check helps make sure openssl doesn't sneak in!


`cargo test` and `cargo clippy` are the things you may want to run locally! The deny don't worry about too much but if it triggers in a PR where you add a crate I'll dig into it with ya!

Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[workspace]
+members = ["db", "cli"]
```
coreyja, line 2: Cargo Workspaces allow you to have multiple crates in the same directory and tie them together a bit! This says that the `db` and `cli` sub-directories here are crates we want to include in our Workspace

Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+resolver = "2"
+
+[workspace.dependencies]
+color-eyre = "0.6.3"
+sqlx = { version = "0.7", default-features = false }
+tokio = { version = "1", features = ["full"] }
+tracing = "0.1.40"
+url = "2.5.0"
+uuid = { version = "1.8.0", features = ["v4"] }
```
coreyja, line 11: These are dependencies that can be used by any crate in the workspace. It helps make sure all the crates we write are using the same versions of dependencies!

Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[workspace.lints.clippy]
+pedantic = "deny"
+missing_errors_doc = "allow"
```
coreyja, line 15: We are turning on `pedantic` mode in Clippy which means it turns on lots more extra lints! If these get annoying we can turn some off, but good to start with all the best practices!

cli/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "cli"
+version = "0.0.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tokio = { workspace = true, features = ["full"] }
+db = { path = "../db" }
```
coreyja, line 10: The `cli` crate depends on the `db` crate, we do that via pathing since they are both in the workspace

cli/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+color-eyre = { workspace = true }
+
+[lints]
+workspace = true
```
coreyja, line 14: Each crate needs to opt-in to the clippy lints that the workspace setup, kinda annoying but 🤷

cli/src/main.rs
```diff
@@ -0,0 +1,8 @@
+#[tokio::main]
+async fn main() -> color_eyre::Result<()> {
+    let _db_pool = db::setup_db_pool().await?;
```
coreyja, line 3: You'll be able to use this `db_pool` with `sqlx` to access the Postgres database! Not used yet in the CLI but should be setup and ready for ya

db/src/lib.rs
```diff
@@ -0,0 +1,96 @@
+pub use sqlx;
+use sqlx::postgres::PgPoolOptions;
+pub use sqlx::PgPool;
+
+#[tracing::instrument(err)]
+pub async fn setup_db_pool() -> color_eyre::Result<PgPool> {
```
coreyja, line 6: This should only needed to be called once in the CLI to setup the Database, and then you can pass around (and clone if needed!) the `PgPool` from there!

But this uses the `DATABASE_URL` env var to connect to the database, and then it runs migrations. The AdvisoryLock stuff is to make sure that only 1 server at a time is trying to run migrations against the DB. Not needed now for local dev and stuff but might be when we deploy to prod so good to setup now!

coreyja, line 6: This should only needed to be called once in the CLI to setup the Database, and then you can pass around (and clone if needed!) the `PgPool` from there!

But this uses the `DATABASE_URL` env var to connect to the database, and then it runs migrations. The AdvisoryLock stuff is to make sure that only 1 server at a time is trying to run migrations against the DB. Not needed now for local dev and stuff but might be when we deploy to prod so good to setup now!

db/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+
+        impl TestPool {
+            async fn teardown(self) -> color_eyre::Result<()> {
+                self.pool.close().await;
+                Postgres::drop_database(&self.database_url).await?;
+                Ok(())
+            }
+        }
+
+        async fn create_test_db_pool() -> color_eyre::Result<TestPool> {
```
coreyja, line 53: This is a helper you can use in tests! Instead of the 'real' `setup_db_pool` in specs we want to use this helper instead!

The difference is that this one will setup a temporary test DB to connect to, and delete that DB when the tests are done. That way different tests don't have a way to interact and leave the DB in a 'dirty' state!

So in a test that needs the DB you can call `create_test_db_pool` and then just remember to call `teardown` at the end to cleanup the test db!
[See example down below!]

db/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+            let database_url = db_url.to_string();
+            Postgres::create_database(&database_url).await?;
+            std::env::set_var("DATABASE_URL", &database_url);
+
+            let pool = setup_db_pool().await?;
+
+            Ok(TestPool { database_url, pool })
+        }
+
+        #[tokio::test]
```
coreyja, line 68: You can use `#[tokio::test]` for tests that need an `async` function instead of a plan function.

I think there is a good change most of your tests end up being async so you'll want this version! Any function with the argument will automatically be run with you do `cargo test`

deny.toml
```diff
@@ -0,0 +1,263 @@
+# This template contains all of the possible sections and their default values
```
coreyja, line 1: This file is really long 😆 

Its basically the empty one the tool generates, and I'll comment on the one place Ive made changes

deny.toml
```diff
@@ -0,0 +1,263 @@
...
+deny = [
+  # Each entry the name of a crate and a version range. If version is
+  # not specified, all versions will be matched.
+  #{ name = "ansi_term", version = "=0.11.0" },
+  #
+  # Wrapper crates can optionally be specified to allow the crate when it
+  # is a direct dependency of the otherwise banned crate
+  #{ name = "ansi_term", version = "=0.11.0", wrappers = [] },
+
+  { name = "openssl" },
```
coreyja, line 207: Here is where we ban `openssl` from the repo!

scripts/setup.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
```
coreyja, line 3: Some scripts to help make this easy to work with!

This one will install the `sqlx-cli` if needed, and then it also creates and migrates the Database

It DOES require the `DATABASE_URL` to be set and for Postgres to be installed

scripts/migrate.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -e
+
+pushd $(git rev-parse --show-toplevel) > /dev/null
+
+cargo sqlx migrate run --source db/migrations
```
coreyja, line 7: This `sqlx migrate` command depends on where you run it from, and I always get annoyed by that 😆 

So this script does NOT care what directory you are in when you run it! So you can run it from in the `db` dir or from the crate root

PR #3, opened 2024-05-22
initilazing  sign up function

Seif-Mamdouh:
This PR allows the us to create users through the `cargo run -- signup` command. 

To sign up a new user:
* run `cargo run -- signup`
* input a username after being prompted by the program
* wait to receive a confirmation log with your `user_name` and `user_id` fields

A couple of thoughts whilst building this:
* it seems like you have to run the `export DATABASE_URL="postgres://localhost:5432/knowledge"` command before being able to run `cargo run` without any errors. I tried adding a `.env` file to the base of the repo, but that didn't seem to work - would love to figure out how to get a `.env` file working so I don't have to run that command every time 😅 

* In order to support an interface that is powered through the CLI, I added a crate called [clap](https://crates.io/crates/clap) - this creates a helpful abstraction for parsing out arguments from cli commands

* To persist the user's authenticated session, we store a local `auth.txt` file that contains the user's `user_id`. This is a temporary work around to building a full fledged auth solution. I'd imagine that as the user makes subsequent requests (after signing up or loggin in), we'd just check the local `auth.txt` file to see who's logged in. 

* _before this PR_ we had no pre-existing database tables, so I wrote a migration and couple of methods that create our initial version of the user table


cli/src/main.rs
```diff
@@ -1,8 +1,89 @@
...
+        Command::Signup => {
+            sign_up(&db_pool).await?;
+        },
+        Command::DisplayUsers => {
+            display_users(&db_pool).await?;
+        }
+    }
+
+    Ok(())
+}
```
Seif-Mamdouh: This is the different types of subcommands available for the users. If we want to add more subcommands. They will go here.

cli/src/main.rs
```diff
@@ -1,8 +1,89 @@
...
+    #[command(subcommand)]
+    command: Command,
+}
+
+#[derive(Subcommand)]
+enum Command {
+    Login,
+    Signup,
+    DisplayUsers,
+}
```
Seif-Mamdouh: Here we have clap [subcommands](https://docs.rs/clap/latest/clap/trait.Subcommand.html). We used this 

`cargo run -- signup --username alice`

in the above example, _signup_ is a subcommand whereas `username` is an argument. I was thinking it'd be nice to call an explicit "signup" command to signify the action you're trying to take, and then collect the arguments by prompting the user:

```
cargo run -- signup
enter a username:
seif
User ID: c750689f-572f-4b0d-b541-621085a74a51, Username: seif
```

An alternate approach would be to collect arguments along the subcommand:

```
cargo run -- signup --username alice
User ID: c750689f-572f-4b0d-b541-621085a74a51, Username: alice
```

OR we can just simplify this heavily and assume the program only runs signup and no other commands
```
cargo run --username alice
User ID: c750689f-572f-4b0d-b541-621085a74a51, Username: alice
```

_but_ the commands were nice for allowing us to check other things, like all the users in the DB (through the `display-users` subcommand for ex.)
```
cargo run -- display-users
User ID: c750689f-572f-4b0d-b541-621085a74a51, Username: alice
User ID: ac849baa-3cb1-40ef-a464-92460fa47617, Username: seif
```

cli/src/main.rs
```diff
@@ -1,8 +1,89 @@
...
+    }
+
+    Ok(())
+}
+
+async fn persist_auth_session(user_id: Uuid) -> color_eyre::Result<()> {
+    if fs::metadata("auth.txt").is_ok() {
+        return Err(color_eyre::eyre::eyre!("auth.txt already exists"));
+    }
+
```
Seif-Mamdouh: We are doing this on purpose to persist user auth session. 

To test it out after you try to sign up in the beginning, sign up with another user name.

```
cargo run -- signup
enter a username:
Corey
Error: auth.txt already exists
```

if you try to sign up with the same username it will say that the user already exists.

```
cargo run -- signup
enter a username:
seif
Error: error returned from database: duplicate key value violates unique constraint "users_user_name_key"

Caused by:
    duplicate key value violates unique constraint "users_user_name_key"
``` 
The message `Error: error returned from database: duplicate key value violates unique constraint "users_user_name_key` is not coming from the persist_auth_session function

cli/src/main.rs
```diff
@@ -1,8 +1,89 @@
...
+    .fetch_one(pool)
+    .await?;
+
+    persist_auth_session(result.user_id).await?;
+
+    println!("User {} added successfully with ID {}!", username, result.user_id);
+    Ok(())
+}
+
+async fn display_users(pool: &PgPool) -> color_eyre::Result<()> {
```
Seif-Mamdouh, line 46: I added this helper method to display the users to check if I users I added to the table are there.

To test 

```
 cargo run -- display-users 

User ID: c750689f-572f-4b0d-b541-621085a74a51, Username: alice
User ID: ac849baa-3cb1-40ef-a464-92460fa47617, Username: seif
User ID: 87f00fd9-41cc-4375-9a15-01ab720b7b98, Username: ali
```

coreyja:
First off this PR Description and all the comments are GREAT!! I'll leave more detailed feedback on specific bits, but overall this is _awesome_!

> it seems like you have to run the export DATABASE_URL="postgres://localhost:5432/knowledge" command before being able to run cargo run without any errors. I tried adding a .env file to the base of the repo, but that didn't seem to work - would love to figure out how to get a .env file working so I don't have to run that command every time 😅

Ya lets fix that 😆 

There are lots on competing solutions but the one that I'm currently using and like is `direnv`: https://direnv.net/
Should be able to do `brew install direnv` and then add something to your shell startup file: https://direnv.net/docs/hook.html

Then when you `cd` into a directory with an `.envrc` file it will load those env vars!

cli/.env
```diff
@@ -0,0 +1 @@
+DATABASE_URL=postgres://localhost:5432/knowledge
```
coreyja: I have my `.envrc` file in repo root, and I have that added to the `.gitignore` too!

That way `git` won't bother you about changes to that file or anything either!

auth.txt
```diff
@@ -0,0 +1 @@
+ac849baa-3cb1-40ef-a464-92460fa47617
```
coreyja: Ooh lets add this `auth.txt` file to the `.gitignore` so that git doesn't pick it up 
(and you might need to run `git rm --cached` to make git 'forget' about this file)

So if you add `cli/auth.txt` as its own line to the `.gitignore` file it will stop picking up changes to this file so each time we login/out we dont get a git diff!

.sqlx/query-1fe4eb033e38d873d0a875093c4bfaea0f8d02c8a70744bc76cb067fee6e3a5c.json
```diff
@@ -0,0 +1,26 @@
+{
```
coreyja, line 1: Ohh!! I like that you generated this file!

Just wanted to explain what it is and why we will want!

`sqlx` is a bit 'special' in that when you compile the `query!` macro it actually connects to the Postgres database and checks what tables and columns exist and all that. Which is really cool cause it can provide us with more type safety!

But the reason we want/need this file is that its kinda annoying to NEED a postgres database to be able to compile this code. On our laptops its not much of an issue, but when we want to deploy this code to a web server, or run tests in CI sometimes its more convenient to not need Postgres setup

So if we set the `SQLX_OFFLINE=1` then sqlx will use this instead of the DB to check the schema.

You already did it once but to type it out this is generated with `cargo sqlx prepare --workspace` (the `--workspace` bit is nice since we have the `db` and `cli` crates)

cli/Cargo.toml
```diff
@@ -9,6 +9,9 @@ edition = "2021"
 tokio = { workspace = true, features = ["full"] }
 db = { path = "../db" }
 color-eyre = { workspace = true }
+clap = { version = "4.1", features = ["derive"] }
+sqlx.workspace = true
+uuid.workspace = true
```
coreyja, line 14: 💯 

Love using the `.workspace` when we can and need/want to share the crate!

cli/src/main.rs
```diff
@@ -1,8 +1,89 @@
+use db::PgPool;
+use tokio;
+use clap::{Parser, Subcommand};
+use db::sqlx;
+use std::fs;
+use std::io::Write;
+use uuid::Uuid; // Ensure you have the `uuid` crate in your Cargo.toml
```
coreyja: SuperNit: Lets cleanup this copy-pasta comment lol

cli/src/main.rs
coreyja: This comment is PERFECT 💯 

I love that not only did you explain what _AND_ why, but you gave alternatives too! 💜 

I like what you have now! The "interactive" sub-command style

But! If its not crazy complicated I think it would be really cool to support the `--username` style too!
Do you think it would work where if you don't give the `--username` option it prompts for it interactively?

cli/src/main.rs
```diff
@@ -1,8 +1,89 @@
...
+
+    Ok(())
+}
+
+async fn persist_auth_session(user_id: Uuid) -> color_eyre::Result<()> {
+    if fs::metadata("auth.txt").is_ok() {
+        return Err(color_eyre::eyre::eyre!("auth.txt already exists"));
+    }
+
+    let mut file = fs::File::create("auth.txt")?;
```
coreyja: Nit: Since you already checked if the file exists it doesn't _really_ matter much, but since our intention is to always make a new file we might want to use `create_new` here instead
https://doc.rust-lang.org/std/fs/struct.File.html#method.create_new

This is cause `create` till truncate the file if one already exists, and `create_new` will error instead

cli/src/main.rs
```diff
@@ -1,8 +1,89 @@
...
+        Command::DisplayUsers => {
+            display_users(&db_pool).await?;
+        }
+    }
+
+    Ok(())
+}
+
+async fn persist_auth_session(user_id: Uuid) -> color_eyre::Result<()> {
+    if fs::metadata("auth.txt").is_ok() {
```
coreyja: So this definitely works! But we can be a bit more specific with this check likely!

I think we might want to use either `Path::exists` or maybe `Path::try_exists`!

These will _just_ check that the file exists, and doesn't grab the metadata.

https://doc.rust-lang.org/std/path/struct.Path.html#method.try_exists
https://doc.rust-lang.org/std/path/struct.Path.html#method.exists

The difference between the two is that `try_exists` will return an `Err` is there are errors checking the file system, while  `exists` will simply return `false`
This allows `try_exists` to be able to tell the difference between an error and the file existing!
But I don't think that we really care about the difference between those two cases here, so I think `exists` is likely fine!

cli/src/main.rs
```diff
@@ -1,8 +1,89 @@
...
+        return Err(color_eyre::eyre::eyre!("auth.txt already exists"));
```
coreyja: Oh I think it would be kinda cool to have this error message tell you who you are currently logged in as! Either the `id` or even the username if its reasonable

What do you think?

cli/src/main.rs
coreyja: This is great! Thanks for showing some examples of the error paths!

> The message Error: error returned from database: duplicate key value violates unique constraint "users_user_name_key is not coming from the persist_auth_session function

It might be nice to look into the `wrap_err` method that comes from `color_eyre` (well really from the `eyre` crate it uses, but close enough lol)
https://docs.rs/eyre/0.6.12/eyre/trait.WrapErr.html#tymethod.wrap_err
https://docs.rs/eyre/0.6.12/eyre/index.html#details

That will let you specify a more "custom" error message to also show the user! Your new "app specific" error message will be shown first, with the original error message being show as 'context' below it!

cli/src/main.rs
coreyja: This is a great little addition, love it!

cli/src/main.rs
```diff
@@ -1,8 +1,89 @@
...
+
+    Ok(())
+}
 
+async fn sign_up(pool: &PgPool) -> color_eyre::Result<()> {
+    println!("Enter a username");
+    let mut username = String::new();
+    std::io::stdin().read_line(&mut username).expect("Failed to read line");
+    let username = username.trim().to_string();
+    add_user(pool, &username).await?;
```
coreyja: `add_user` here takes a `&str`, which is what `trim()` returns! So you can probably remove the `to_string()` here!

The difference is that a `&str` is a reference to an existing piece of memory, and by calling `to_string` you are making an Owned String type, which needs a new bit of memory

So be leaving off the `to_string` we can save a small memory allocation!

Now if the compiler is mad at you with the reference, and this was to make it happy then never mind 😆 
Rust gets the most complicated with references, and thats where the infamous borrow checker comes in and gives lots of people problems. Cloning or `to_string`ing can get out of those problems and the cost of slightly worse theoritical performance. But OFTEN its not really worth it to fight the compiler, and a little `clone` or `to_string` just makes this easier lol

All that to say, if you can remove the `to_string` and it compiles lets do it! BUT if the compiler yells at ya, I'm 100% fine leaving this in lol!
Its not usually worth it to 'over optimize' things like this until we know its a slow code path or something

db/Cargo.toml
```diff
@@ -6,15 +6,8 @@ edition = "2021"
...
-  "postgres",
-  "runtime-tokio-rustls",
-  "chrono",
-  "migrate",
-  "macros",
-  "uuid",
-] }
-color-eyre = { workspace = true }
+sqlx = { workspace = true, features = ["postgres", "runtime-tokio-rustls", "chrono", "migrate", "macros", "uuid"] }
+color-eyre = "0.6.3"
```
coreyja: Lets use the `color-eyre = { workspace = true }` version!

That way we only have a single version of `color-eyre` in the workspace!

db/migrations/20240523181636_userTable.sql
```diff
@@ -0,0 +1,7 @@
+-- Ensure the UUID extension is enabled
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+
+CREATE TABLE Users (
+  user_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
```
coreyja, line 5: This schema is perfect!

Just wanted to call out this line, and why I prefer this style of `user_id` (as opposed to just `id`)!

I've often done an `id` here, and then when we need to refer to it in a different model used `user_id` there. And this _works_ but it makes one specific thing annoying! When you need to do a SQL join its more annoying to type the relation.

Lets assume we have a Comments table just as an example. To join Users and Comments the `id` and `user_id` way would look like this:
```
JOIN Comments on Comments.user_id = Users.id
```

And if we have `user_id` on both side we can use this `using` keyword!

```
JOIN Comments USING (user_id)
```

Which is definitely not the biggest deal, but I like being able to use the `using` syntax!

db/src/lib.rs
```diff
@@ -1,6 +1,27 @@
 pub use sqlx;
 use sqlx::postgres::PgPoolOptions;
 pub use sqlx::PgPool;
+use color_eyre::Result;
+
+
+#[derive(sqlx::FromRow)]
+pub struct User {
+    pub user_id: String,
```
coreyja: Oh I think I told you to use a `String` here, so sorry I'm about to tell you to change it 😆 

But I think we can use the `Uuid` type from the `uuid` crate here!

That way its clearer from just the Rust code that we are storing a `uuid` here as opposed to just any arbitrary string!

db/src/lib.rs
```diff
@@ -11,9 +32,10 @@ pub async fn setup_db_pool() -> color_eyre::Result<PgPool> {
         .await?;
     let mut connection = pool.acquire().await?;
 
-    let lock = sqlx::postgres::PgAdvisoryLock::new("knowledge-db-migration-lock");
+    let lock = sqlx::postgres::PgAdvisoryLock::new("running the show!");
```
coreyja: 😆 

Lets revert this back I think too! But I do like your message lol

db/src/lib.rs
```diff
@@ -11,9 +32,10 @@ pub async fn setup_db_pool() -> color_eyre::Result<PgPool> {
...
     let mut lock = lock.acquire(&mut connection).await?;
 
+    // sqlx::migrate!().run(connection.as_mut()).await?;
```
coreyja: SuperNit: We can clean up this line

cli/src/main.rs
```diff
@@ -1,8 +1,89 @@
...
+    Ok(())
+}
+
+async fn add_user(pool: &PgPool, username: &str) -> color_eyre::Result<()> {
+    let result = sqlx::query!(
+        "INSERT INTO Users (user_name) VALUES ($1) RETURNING user_id",
+        username
+    )
+    .fetch_one(pool)
+    .await?;
```
coreyja: What do you think about moving this to the `db` create maybe as a `create_user` method and then calling that here?

That way we can isolate all the DB interations to the DB crate!
One nice thing that I think will be easier if we do that, is switching out this direct DB interaction for an API eventually.

The CLI talking to the DB 'directly' is great for getting started today, but if we were to ship this CLI to users we would NOT want them to be able to write to the DB directly, cause that would be chaos lol!
Instead we'd have our server write to the Database, based on API requests from the cli clients.

And I think that transition will be easier if we can "just" remove the `db` crate from the CLI and fill in all the gaps with API calls lol

coreyja:
This is so good!! Left a lot of comments, a good few of them just chatting!

I also pushed one more commit to do some Github Actions setup to get tests running!

I _think_ both the red checks now are 'real' issues! All of the issues from those _should_ be included as Github Comments in the "Files" tab!

And you can do the two bits locally too to test out!

The `clippy` one is just running `cargo clippy --all-targets`
And the `tests` is running `cargo test`

cli/src/main.rs
Seif-Mamdouh: I just added a commit to sign-up functionality now supports both

` cargo run -- signup 
   Enter username: 
`

and 

`cargo run -- signup --username <Enter Username>`

.sqlx/query-1fe4eb033e38d873d0a875093c4bfaea0f8d02c8a70744bc76cb067fee6e3a5c.json
Seif-Mamdouh: Awesome! That's very cool!

cli/src/main.rs
Seif-Mamdouh: I think it's great idea, that way the user would know if they are logged in already or not

db/migrations/20240523181636_userTable.sql
Seif-Mamdouh: Awesome! Thank you!

cli/src/main.rs
Seif-Mamdouh: My compiler is yelling at me 😆

Seif-Mamdouh: I tried removing it but the complier just wasn't having it

db/src/lib.rs
```diff
@@ -7,7 +7,7 @@ use uuid::Uuid;
 
 #[derive(sqlx::FromRow)]
 pub struct User {
-    pub user_id: String,
+    pub user_id: Uuid,
```
Seif-Mamdouh, line 10: added Uuid ad Data type instead of string

cli/src/main.rs
```diff
@@ -16,7 +16,10 @@ struct KnowledgeArgs {
 #[derive(Subcommand)]
 enum Command {
     Login,
-    Signup,
+    Signup {
+        #[arg(short, long)]
+        username: Option<String>,
+    },
```
Seif-Mamdouh, line 22: I added another attribute here to take in the username subcommand

.gitignore
```diff
@@ -1,3 +1,7 @@
 .envrc
 
 target/
+
+auth.txt
+
+.envrc
```
coreyja: SuperNits: First lets add a new line to this file! In Github you can see its missing one cause of this symbol

image

And then second we should be able to remove this `.envrc` since there is one already up top on line 1!

cli/src/main.rs
coreyja: Perfect!

cli/src/main.rs
```diff
@@ -1,8 +1,100 @@
...
+}
+
+async fn sign_up(pool: &PgPool, username_opt: Option<String>) -> color_eyre::Result<()> {
+    if Path::new("auth.txt").exists() {
+        let mut file = fs::File::open("auth.txt")?;
+        let mut contents = String::new();
+        file.read_to_string(&mut contents)?;
+        let user_id = Uuid::parse_str(contents.trim())?;
+
+        let user = db::get_username_by_id(pool, user_id).await?;
```
coreyja: ❤️

cli/src/main.rs
```diff
@@ -1,8 +1,100 @@
...
+        return Err(color_eyre::eyre::eyre!("You are signed up as {}", user)
+            .wrap_err("Signup attempt failed due to existing user session"));
```
coreyja: Nice! I like this!

I wouldn't have thought to use a "real" error message and then provide more details as 'context' on what went wrong with the `wrap_err`! But I like how this is going to display 💯

cli/src/main.rs
```diff
@@ -1,8 +1,100 @@
...
+
+    let username = match username_opt {
+        Some(name) => name,
+        None => {
+            println!("Enter a username:");
+            let mut input = String::new();
+            std::io::stdin().read_line(&mut input).expect("Failed to read line");
+            input.trim().to_string()
+        }
+    };
```
coreyja: Love it!

cli/src/main.rs
```diff
@@ -1,8 +1,100 @@
...
+async fn add_user(pool: &PgPool, username: &str) -> color_eyre::Result<()> {
+    let user_id = db::create_user(pool, username).await?;
+    persist_auth_session(pool, user_id).await?;
+
+    println!("User {} added successfully with ID {}!", username, user_id);
+    Ok(())
+}
 
-    println!("Hello, world!");
+async fn persist_auth_session(pool: &PgPool, user_id: Uuid) -> color_eyre::Result<()> {
```
coreyja: I think we can remove the `pool` here, and I think this is one of the clippy lints!

Since we aren't using `pool` here we can remove it from the args here

db/Cargo.toml
```diff
@@ -6,15 +6,8 @@ edition = "2021"
...
-sqlx = { workspace = true, features = [
-  "postgres",
-  "runtime-tokio-rustls",
-  "chrono",
-  "migrate",
-  "macros",
-  "uuid",
-] }
 color-eyre = { workspace = true }
+sqlx = { workspace = true, features = ["postgres", "runtime-tokio-rustls", "chrono", "migrate", "macros", "uuid"] }
```
coreyja, line 10: Supernit: I kinda like the 'old' version with the new lines just for readability but really not a big deal lol

And I think this list of features is the same in your version right? Seems to be

db/src/lib.rs
```diff
@@ -1,6 +1,51 @@
...
+
+
+#[derive(sqlx::FromRow)]
+pub struct User {
+    pub user_id: Uuid,
+    pub user_name: String,
+}
+
+
+pub async fn create_user(pool: &PgPool, username: &str) -> color_eyre::Result<Uuid> {
```
coreyja: Oh this `clippy` is interesting 🤔 

I'll likely push a commit to turn this off, at least for now!

The clippy is saying

> docs for function which may panic missing `# Panics` section

The idea here is that in Rust there are kinda two ways to do errors. The 'right' way with returning a `Result` like we are going, or you can 'panic' with `unwrap` or `panic!` or something like that.
The lint wants you to add a comment above this function that explains why this function might panic as a way to let other developers know a panic is possible in this function

BUTT I don't _see_ what could panic here.. Maybe `query!` can panic or something... But I don't think we want/need to care about that panic here, so probably makes sense to ignore this one

I'll push a commit to this PR in like 5 min to do that!

coreyja: Ohhb 🤔  is `fetch_one` what is causing the panic here? Ahh ya I think it is!

Ok so maybe we do want to deal with this! I think we want to look at the times you used `fetch_one` and see if it makes sense to use `fetch_one` and write a panic comment or if we should switch to using `fetch_optional` (which returns an Option<T> instead of a T)

In this case I think the `fetch_one` is great, because I don't think there is any reasonable case where this query wouldn't return any results (since its creating the record). We may get an error about the username existing, but I think that will come as a different `Err` anyways, so I think panicing here is fine!

So lets add a `doc comment` https://doc.rust-lang.org/reference/comments.html#doc-comments
Above this function with a `# Panics` section, that says this function could panic if the `INSERT` row doesn't return a row, but we don't expect that to happen in any normal circumstances.

And then with that `clippy` should be happy here!

db/src/lib.rs
```diff
@@ -1,6 +1,51 @@
...
+         " 
+         SELECT user_id, user_name 
+         FROM Users
+         ",
+    ).fetch_all(pool).await?;
+
+    Ok(users)
+}
+
+pub async fn get_username_by_id(pool: &PgPool, user_id: Uuid) -> color_eyre::Result<String> {
```
coreyja: In this function I think we want to use `fetch_optional` instead of `fetch_one` for the case where the `id` doesn't exist in the DB

We'll probably want to return a `Result<Option<String>>` here since there is a chance the Username can't be found

And I think this will also fix the clippy lint here!

db/src/lib.rs
Seif-Mamdouh: I did add comments I just never committed it because I wasn't sure I was correct
Screenshot 2024-05-28 at 11 15 20 AM

Seif-Mamdouh: should I comment this?

cli/src/main.rs
```diff
@@ -63,9 +63,13 @@ async fn sign_up(pool: &PgPool, username_opt: Option<String>) -> color_eyre::Res
...
-        let user = db::get_username_by_id(pool, user_id).await?;
-        return Err(color_eyre::eyre::eyre!("You are signed up as {}", user)
-            .wrap_err("Signup attempt failed due to existing user session"));
+        match db::get_username_by_id(pool, user_id).await? {
+            Some(existing_username) => {
+                println!("User ID already registered with username: {existing_username}");
+                return Ok(());
+            },
+            None => println!("No existing user found with this ID, proceeding with signup."),
+        }
```
Seif-Mamdouh, line 72: used match to handle the optional Result string from get_username_by_id

db/src/lib.rs
```diff
@@ -12,10 +12,24 @@ pub struct User {
 }
 
 
-pub async fn create_user(pool: &PgPool, username: &str) -> color_eyre::Result<Uuid> {
+/// Creates a new user with the given username.
```
Seif-Mamdouh, line 15: added the panic comments here to give it a shot. Let me know if I did this correctly

db/src/lib.rs
```diff
@@ -35,15 +49,15 @@ pub async fn get_users(pool: &PgPool) -> Result<Vec<User>> {
...
     let record = sqlx::query!(
         "SELECT user_name FROM Users WHERE user_id = $1",
         user_id
     )
-    .fetch_one(pool)
+    .fetch_optional(pool)
     .await?;
 
-    Ok(record.user_name)
+    Ok(record.map(|r| r.user_name))
```
Seif-Mamdouh, line 60: switched from fetch_one to fetch_optional 

handleed the result in the Ok() clause

db/src/lib.rs
coreyja: This is PERFECT!!

cli/src/main.rs
```diff
@@ -1,10 +1,10 @@
-use db::PgPool;
 use clap::{Parser, Subcommand};
+use db::PgPool;
 use std::fs;
+use std::io::Read;
 use std::io::Write;
-use uuid::Uuid;
 use std::path::Path;
-use std::io::Read;
+use uuid::Uuid;
```
coreyja: Oh! My editor runs `cargo fmt` by default and this re-arranged some of these, and removed some commas down below and all that!

We'll probably wire up fmt to CI to make sure its run, but we don't need to worry about that now!

cli/src/main.rs
```diff
@@ -88,13 +90,13 @@ async fn sign_up(pool: &PgPool, username_opt: Option<String>) -> color_eyre::Res
...
 async fn add_user(pool: &PgPool, username: &str) -> color_eyre::Result<()> {
     let user_id = db::create_user(pool, username).await?;
-    persist_auth_session(user_id).await?;
+    persist_auth_session(user_id)?;
 
     println!("User {username} added successfully with ID {user_id}!");
     Ok(())
 }
 
-async fn persist_auth_session(user_id: Uuid) -> color_eyre::Result<()> {
```
coreyja: This was the 'last' clippy that I ran into when I turned off that panic lint

It was saying that this async wasn't needed since there were no `awaits` in the function! So removed it and removed the `.await` from where we call this function

db/src/lib.rs
coreyja: Oh sorry I didn't see this Github Comment but ya I love this! 💯

.gitignore
coreyja: Ahh can we fix this up quick before we merge the PR? Not a huge deal tho

coreyja (dismissed):
This is great!!

Everything came out really well, and it definitely seems like you are picking all this up super fast!

I left one more comment on the `.gitignore` file cleanup but after that feel free to merge this PR! 🎉

coreyja (approved):

PR #4, opened 2024-05-28
linear 275 - implement check status functionality

Seif-Mamdouh:
This pr allows users to check status through `cargo run -- status`

**Usage:**
run 
`cargo run -- status
`
Output:
`Logged in with Username: gedo, User ID: a6007aaa-06c1-4282-a23d-099ae418b426`

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -93,7 +93,7 @@ async fn check_status(pool: &PgPool) -> color_eyre::Result<()> {
 
         match db::get_username_by_id(pool, user_id).await? {
             Some(username) => println!("Logged in with Username: {username}, User ID: {user_id}"),
-            None => println!("No user found with User ID: {}", user_id),
+            None => println!("No user found with User ID: {user_id}"),
         }
     } else {
         println!("Not logged in.");
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -93,7 +93,7 @@ async fn check_status(pool: &PgPool) -> color_eyre::Result<()> {
 
         match db::get_username_by_id(pool, user_id).await? {
             Some(username) => println!("Logged in with Username: {username}, User ID: {user_id}"),
-            None => println!("No user found with User ID: {user_id}"),
+            None => println!("file empty"),
         }
     } else {
         println!("Not logged in.");
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -40,7 +40,7 @@ async fn main() -> color_eyre::Result<()> {
             display_users(&db_pool).await?;
         }
         Command::Status => {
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -21,6 +21,7 @@ enum Command {
         username: Option<String>,
     },
     DisplayUsers,
+    Status,
```
Seif-Mamdouh, line 24: Added a subcommand enum here

cli/src/main.rs
```diff
@@ -56,20 +60,11 @@ async fn display_users(pool: &PgPool) -> color_eyre::Result<()> {
...
-
-        match db::get_username_by_id(pool, user_id).await? {
-            Some(existing_username) => {
-                println!("User ID already registered with username: {existing_username}");
-                return Ok(());
-            }
-            None => println!("No existing user found with this ID, proceeding with signup."),
-        }
+        check_status(pool).await?;
+        return Ok(());
```
Seif-Mamdouh, line 67: To avoid repeating the same logic, I called the function `check_status` here

cli/src/main.rs
```diff
@@ -88,6 +83,24 @@ async fn sign_up(pool: &PgPool, username_opt: Option<String>) -> color_eyre::Res
...
+        match db::get_username_by_id(pool, user_id).await? {
+            Some(username) => println!("Logged in with Username: {username}, User ID: {user_id}"),
+            None => println!("No user found with User ID: {user_id}"),
+        }
+    } else {
+        println!("Not logged in.");
+    }
+    Ok(())
+}
+
```
Seif-Mamdouh, line 103: This function checks for the auth.txt file in the root.
If there is 
it reads into the file and looks for the user_id in the DB to print out the user_id and user_name


`Logged in with Username: *username*, User ID: *user_id*`

if not 
It outputs
`not logged in`

cli/src/main.rs
coreyja: Love that!

cli/src/main.rs
```diff
@@ -88,6 +83,24 @@ async fn sign_up(pool: &PgPool, username_opt: Option<String>) -> color_eyre::Res
     Ok(())
 }
 
+async fn check_status(pool: &PgPool) -> color_eyre::Result<()> {
```
coreyja, line 86: Nit: Maybe this could be `check_auth_status`?

coreyja (approved):
This is great! Pulling out that function was a great call!

I really don't have much else to say here!

cli/src/main.rs
Seif-Mamdouh: will do

PR #5, opened 2024-05-29
Run Cargo fmt and setup CI to verify it

coreyja:
`cargo fmt` can be used to format your code and keeps most Rust code looking relatively consistent which is nice!

This formats the existing code and setups up a CI job to enforce the format in PRs

In VS Code you can configure it to run format on save, so that this happens automatically and you don't really need to think about it at all!

https://stackoverflow.com/a/72282191

### DIFF
--- a/db/src/lib.rs
+++ b/db/src/lib.rs
@@ -1,16 +1,14 @@
+use color_eyre::Result;
 pub use sqlx;
 use sqlx::postgres::PgPoolOptions;
 pub use sqlx::PgPool;
-use color_eyre::Result;
 use uuid::Uuid;
-
 
 #[derive(sqlx::FromRow)]
 pub struct User {
     pub user_id: Uuid,
     pub user_name: String,
 }
-
 
 /// Creates a new user with the given username.
 ///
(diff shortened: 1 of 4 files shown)


deny.toml
```diff
@@ -109,7 +109,7 @@ deny = [
   #"Nokia",
 ]
 # Lint level for licenses considered copyleft
-copyleft = "warn"
```
coreyja, line 112: Oh I snuck in small change here too!

Making this tool help make sure we don't accidentally install copyleft libraries that might force us to open source this code. 

Its 'public' so its more "source available" right now. But I've marked the copyright here as all rights reserved

deny.toml
```diff
@@ -133,6 +133,7 @@ exceptions = [
   # Each entry is the crate and version constraint, and its specific allow
   # list
   #{ allow = ["Zlib"], name = "adler32", version = "*" },
+  { allow = ["MPL-2.0"], name = "webpki-roots", version = "*" },
```
coreyja, line 136: This crate uses the `MPL-2.0` license, and thats considered copy-left but we are allowed to distribute executables with it still so its going to be fine for any of our use cases

db/src/lib.rs
```diff
@@ -40,27 +38,28 @@ pub async fn create_user(pool: &PgPool, user_name: &str) -> color_eyre::Result<U
...
-    )
-    .fetch_optional(pool)
-    .await?;
+pub async fn get_username_by_id(
+    pool: &PgPool,
+    user_id: Uuid,
+) -> color_eyre::Result<Option<String>> {
+    let record = sqlx::query!("SELECT user_name FROM Users WHERE user_id = $1", user_id)
+        .fetch_optional(pool)
+        .await?;
```
Seif-Mamdouh, line 58: I see it formats the code in a certain way. That's nice!

Seif-Mamdouh:
Looks great! `cargo fmt` is a great library to format the code it makes it look cleaner! 

I read https://jimbobbennett.dev/blogs/cargo-fmt/ which helped me understand the full picture.

Seif-Mamdouh (dismissed):

PR #6, opened 2024-05-29
created a new migration and user can add URL now

Seif-Mamdouh:
This pr allows the use to add URL through 
`cargo run add-url --url  <insert-url>`
&& the user can a existing URL through 
`cargo run add-url --url  <insert-url> --allow-existing `

If a user is signed in 
they use either commands 
`cargo run add-url --url  <insert-url>`
`cargo run add-url --url  <insert-url> --allow-existing `

if the user is not logged in 
it will error out


cli/src/add_url.rs
```diff
@@ -0,0 +1,18 @@
+// File: add_url.rs
+use db::PgPool;
+use color_eyre::Result;
+
+use crate::check_auth_status;
+
+pub async fn add_url(pool: &PgPool, url: &str, allow_existing: bool) -> Result<()> {
+
+    check_auth_status(pool).await?;
```
Seif-Mamdouh, line 14: if user is not logged in it will error out. Otherwise the system will proceed

cli/src/add_url.rs
```diff
@@ -0,0 +1,18 @@
...
+
+    let result = db::add_url(pool, url, &allow_existing).await;
+    match result {
+        Ok(message) if message.contains("re-adding is allowed") => println!("{message}"),
+        Ok(message) => println!("URL added successfully with ID: {message}"),
+        Err(e) => println!("Error: {e}"),
+    }
+    Ok(())
+}
```
Seif-Mamdouh: this function call the add_url from the DB that is used in the result variable to match certain outputs for the user.

In this section if the user enter a url that is not in the DB however they pass `-- allow existing`. The system will recognize that the URL wasn't present and automatically add it. 

The output would display `URL added successfully with ID: <url>`

We could go into more depth and output a different output such as 
`Url  wasn't present in the DB, adding it now..... You can now use -- allow-existing `

Let me know what you think

cli/src/check_auth_status.rs
```diff
@@ -0,0 +1,24 @@
...
+        match db::get_username_by_id(pool, user_id).await? {
+            Some(username) => println!("Logged in with Username: {username}, User ID: {user_id}"),
+            None => println!("file empty"),
+        }
+    } else {
+        return Err(color_eyre::eyre::eyre!("Not logged in."));
+    }
+    Ok(())
+
+}
```
Seif-Mamdouh: Did some refactoring here. I just moved it into a separate file. Let me know if you like this style or is there a specific style you would like me to do 🤔

cli/src/main.rs
```diff
@@ -62,7 +77,7 @@ async fn display_users(pool: &PgPool) -> color_eyre::Result<()> {
 
 async fn sign_up(pool: &PgPool, username_opt: Option<String>) -> color_eyre::Result<()> {
     if Path::new("auth.txt").exists() {
-        check_status(pool).await?;
+        check_auth_status(pool).await?;
```
Seif-Mamdouh, line 82: changed the name as suggested in the previous PR

db/migrations/20240529151056_Page.sql
```diff
@@ -0,0 +1,9 @@
+-- Ensure the UUID extension is enabled
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+
+CREATE TABLE Page (
+    page_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    user_id Uuid UNIQUE,
+    url_id Uuid UNIQUE,
+    url VARCHAR(255) NOT NULL
+);
```
Seif-Mamdouh: created a new migration table

db/src/lib.rs
```diff
@@ -10,20 +11,16 @@ pub struct User {
...
+#[derive(sqlx::FromRow)]
+struct UrlRecord {
+    url_id: Option<Uuid>,
+}
+
+#[derive(sqlx::FromRow)]
+struct ExistRecord {
+    exists: Option<bool>,
+}
+
```
Seif-Mamdouh, line 26: This struct is used for deserialization for primitive data types such as Bool that is used in `sqlx::query_as!`

db/src/lib.rs
```diff
@@ -35,6 +32,34 @@ pub async fn create_user(pool: &PgPool, user_name: &str) -> color_eyre::Result<U
...
 }
 
+pub async fn add_url(pool: &PgPool, url: &str, allow_existing: &bool) -> color_eyre::Result<String> {
+    let exist_record = sqlx::query_as!(
+        ExistRecord,
+        "SELECT EXISTS(SELECT 1 FROM Page WHERE url = $1) as exists",
+        url
+    )
+    .fetch_one(pool)
+    .await?;
```
Seif-Mamdouh: The `exist_record` variable retrieves a boolean value indicating whether the specified URL already exists in the `Page` table. 

The way we determine if it exisits is by the SQL query

The result, `true` or `false`, is encapsulated within the `ExistRecord` struct. 

This check is crucial for determining the next appropriate action based on the `allow_existing` flag provided by the user.

db/src/lib.rs
```diff
@@ -35,6 +32,34 @@ pub async fn create_user(pool: &PgPool, user_name: &str) -> color_eyre::Result<U
...
+
+    // exist_record might be true or false
+    if let Some(true) = exist_record.exists {
+        if *allow_existing {
```
Seif-Mamdouh: In this code, I used the dereferencing operator ` *` to access the underlying value of the `allow_existing ` reference. This is necessary because I need to evaluate the actual boolean value in the conditional expression.

cli/src/check_auth_status.rs
coreyja: I like this!

I think instead of naming the module/file the same as the function maybe we can group by 'domain' a bit more?
So like maybe this is the `auth` module/file? and we can add more auth related stuff here as it exists?

No strong feels tho! Easy to move around later too lol

cli/src/check_auth_status.rs
```diff
@@ -0,0 +1,24 @@
...
+    let user_path = Path::new("auth.txt");
+    if user_path.exists() {
+        let mut file = fs::File::open("auth.txt")?;
+        let mut contents = String::new();
+        file.read_to_string(&mut contents)?;
+        let user_id = Uuid::parse_str(contents.trim())?;
+
+        match db::get_username_by_id(pool, user_id).await? {
+            Some(username) => println!("Logged in with Username: {username}, User ID: {user_id}"),
+            None => println!("file empty"),
```
coreyja: Should this case return an error too?

Cause right now it would print this out, but then end up returning the `Ok(())` down below right?

db/migrations/20240523181636_userTable.sql
```diff
@@ -4,4 +4,4 @@ CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
 CREATE TABLE Users (
   user_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
   user_name VARCHAR(255) NOT NULL UNIQUE
-);
+);
```
coreyja: Ahh I think this might be why you had that error about an existing migration having changed!

And in general I think we want the new line at the end of this file so probably good to change this back.
But that might make you drop and re-create the DB again since this migration hash will change again lol

db/migrations/20240529151056_Page.sql
```diff
@@ -0,0 +1,9 @@
+-- Ensure the UUID extension is enabled
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+
+CREATE TABLE Page (
+    page_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    user_id Uuid UNIQUE,
+    url_id Uuid UNIQUE,
```
coreyja: Do we need this `url_id` field?

I typically only add `x_id` columns when we have a table for whatever it is. So since `url` is 'just' a column on `Page` and not a table itself I don't think we'll need this `url_id`!

db/migrations/20240529151056_Page.sql
```diff
@@ -0,0 +1,9 @@
+-- Ensure the UUID extension is enabled
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+
+CREATE TABLE Page (
+    page_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    user_id Uuid UNIQUE,
```
coreyja: I think we probably do NOT want to make this unique, or else I believe a User will only be able to add a single Page to the DB!

And I also think we want to add a Foreign Key Contraint here! I believe you can do that by adding the FK constraint with a `REFERENCES` after this column!

```
user_id UUID REFERENCES Users (user_id)
```

What that does is have Postgres make sure we get the `uuids` right! So if we tried to create a Page and gave it a `user_id` that doesn't exist in `Users#user_id` it will error out for us!

coreyja: And we probably want to add `NOT NULL` here too, to make sure that every page is owned by a User

db/migrations/20240529151056_Page.sql
coreyja: Nit: Can we add a newline at the end of this file?

Here is a way to make VS Code do this automatically so you don't have to worry about it!
https://stackoverflow.com/a/44704969

db/src/lib.rs
coreyja: Totally makes sense!

In cases like this where you need to do a SQL query, and just need the result within your function you might find it easier to use `query!` as opposed to `query_as!`

When you use `query!` you get a 'Record' back that magically has the columns and types that the SQL returned!
The downside is that its not a 'named struct' so you can't really pass it around nicely. But for something like this where you just need to use the boolean I think it might work nicely!

db/src/lib.rs
```diff
@@ -35,6 +32,34 @@ pub async fn create_user(pool: &PgPool, user_name: &str) -> color_eyre::Result<U
...
+    } 
+    
+    let result = sqlx::query!(
+        "INSERT INTO Page (url) VALUES ($1) RETURNING url",
+        url
+    )
+    .fetch_one(pool)
+    .await?;
+
+    Ok(result.url)
```
coreyja: Nit: This is obviously perfect for this use case!

But forward thinking a bit, I wonder if these 'create' method should/could return the full object on create?

Like we could create a `Page` struct with fields that match the table and return a 'full' Page struct here! That way the consumer could get either the url or the ids depending on what they need

db/src/lib.rs
```diff
@@ -35,6 +32,34 @@ pub async fn create_user(pool: &PgPool, user_name: &str) -> color_eyre::Result<U
     Ok(result.user_id)
 }
 
+pub async fn add_url(pool: &PgPool, url: &str, allow_existing: &bool) -> color_eyre::Result<String> {
+    let exist_record = sqlx::query_as!(
+        ExistRecord,
+        "SELECT EXISTS(SELECT 1 FROM Page WHERE url = $1) as exists",
```
coreyja: This totally works!


I _might_ have written this as something more like 
```
SELECT page_id from Pages WHERE url = $1
```

and then use `fetch_optional` to get the result back as an option. And then `Some(_)` would mean the record exists and `None` would mean its missing!

But I don't see a real reason to change this! Just chatting!

db/migrations/20240529151056_Page.sql
```diff
@@ -0,0 +1,9 @@
+-- Ensure the UUID extension is enabled
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+
+CREATE TABLE Page (
```
coreyja: Can we call the table `Pages`? I think we'll do plural names for all our tables just as a convention

db/migrations/20240529151056_Page.sql
```diff
@@ -0,0 +1,9 @@
...
+    page_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    user_id Uuid UNIQUE,
+    url_id Uuid UNIQUE,
+    url VARCHAR(255) NOT NULL
```
coreyja: Another thing I think we want to add to this table/migration is a UNIQUE index over the `user_id` and `url` columns

That way Postgres can help make sure that for any user they don't have multiple `Pages` that point to the same URL.
But it will allow different users to point save the same URL without issues!

Happy to help with the syntax for this if you run into issues!

db/.sqlx/query-1fe4eb033e38d873d0a875093c4bfaea0f8d02c8a70744bc76cb067fee6e3a5c.json
```diff
@@ -0,0 +1,26 @@
+{
```
coreyja: I think we probably can/want to remove this files in `db/.sqlx` and 'prefer' the one at the root level in `.sqlx`!

I think the different is these ones were likely generated without the `--workspace` option!

If we use the `--workspace` option it should look in all the crates for put the file in the workspace root

Since we are doing all the sql calls in this DB crate it matters less, but we started with the workspace version so likely good to continue with that!

db/src/lib.rs
```diff
@@ -35,6 +32,34 @@ pub async fn create_user(pool: &PgPool, user_name: &str) -> color_eyre::Result<U
     Ok(result.user_id)
 }
 
+pub async fn add_url(pool: &PgPool, url: &str, allow_existing: &bool) -> color_eyre::Result<String> {
```
coreyja: Your implementation is great! And exactly what I would have expected!

Butt looking at with 20 years of experience I think we can optimize it! And honestly it probably doesn't _need_ to be optimized now, and I'm going to break my rule of not optimizing Rust until we know we need it. BUTT I want to chat about this and share 😆 !

Right now this requires two SQL calls to add the URL, one to check if it exists and one to actually add it if needed. And I THINK we can drop this down to a single SQL call in all cases 🎉 

But it is going to be a bit weird/different so please ask questions if it doesn't make sense!

The basic idea is that we can do what is called an `upsert` to either insert or update the record in a single SQL call, and then after that we can tell if the record was just created or not! This 'upsert' will also rely on the unique index on `(user_id, url)` that I mentioned in another comment!

The idea would be something like this code!

```rust
let new_page_id = Uuid::gen_v4();

let upsert_result = sqlx::query_as!(
Pages,
"INSERT INTO Page (page_id, user_id, url) VALUES ($1, $2, $3) RETURNING * ON CONFLICT DO NOTHING",
new_page_id,
user_id,
url)
.fetch_one(pool)
.execute();

let exists = upsert_result.page_id != new_page_id;
```

The basic idea is that we can create a new `page_id` in Rust (instead of letting PG make one for us) and then try to insert the URL with that `page_id` and with `ON CONFLICT DO NOTHING` at the end of the insert. This tells Postgres that if it gets a conflict on a unique index we want it to 'DO NOTHING' or not update any columns.

This means that if we inserted a new Page it will have the `page_id` we specified. But if this User already added this url it will have a _DIFFERENT_ if since the record already existed.

We can check this after the insert to know if it was a new record or not, and error out if the allow_existing flag wasn't passed in!

cli/src/add_url.rs
```diff
@@ -0,0 +1,18 @@
...
+use crate::check_auth_status;
+
+pub async fn add_url(pool: &PgPool, url: &str, allow_existing: bool) -> Result<()> {
+
+    check_auth_status(pool).await?;
+
+    let result = db::add_url(pool, url, &allow_existing).await;
+    match result {
+        Ok(message) if message.contains("re-adding is allowed") => println!("{message}"),
+        Ok(message) => println!("URL added successfully with ID: {message}"),
```
coreyja: Ahh! Here might be a fun case where we want to return something different than just a `String` message so that we can tell these two cases apart!

The easiest answer is maybe we want to return a new Enum instead! We can make a new Enum, maybe called `AddUrlOutcome` (for lack of a more original name lol) with two variants, one for when we created a new URL and one for where an existing one existed!

```
enum AddUrlOutcome {
  Created(String),
  Existing(String),
}
```

And then you can match on the enum more specifically instead of having do a `message.contains` check on the returned message!

cli/src/add_url.rs
coreyja: I like the general output you have now! I don't think it needs to be more in depth really, but I like the suggestion!

coreyja (changes requested):
Looking really good! Left a few more comments this time, mostly on the DB interactions!

Noticed that there are some cargo fmt issues here, so wanted to plug this VS Code setting again that will do fmt on save for you, so you get the formatted results all the time!
It's _really_ nice and now I can't imagine coding without an auto-formatter 😆 

> In VS Code you can configure it to run format on save, so that this happens automatically and you don't really need to think about it at all!
> 
> [stackoverflow.com/a/72282191](https://stackoverflow.com/a/72282191)

db/src/lib.rs
Seif-Mamdouh: Love the suggestion

db/src/lib.rs
Seif-Mamdouh: It was journey but I enjoyed it

db/src/lib.rs
```diff
@@ -44,7 +44,9 @@ pub async fn add_url(
 
     let upsert_result = sqlx::query_as!(
         Page,
-        "INSERT INTO Pages (page_id, user_id, url) VALUES ($1, $2, $3) ON CONFLICT (user_id, url) DO NOTHING RETURNING *",
+        "INSERT INTO Pages (page_id, user_id, url) VALUES ($1, $2, $3)
+         ON CONFLICT (user_id, url) DO UPDATE SET url = EXCLUDED.url
+         RETURNING *",
```
Seif-Mamdouh, line 50: The query will update the existing row that it found with user_id and url.

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "editor.formatOnSave": true,
+    "rust-analyzer.rustfmt.enableRangeFormatting": false,
+    "[rust]": {
+        "editor.defaultFormatter": "matklad.rust-analyzer"
+    },
+    "files.insertFinalNewline": true
+}
```
coreyja, line 8: Oh ya I forget about project specific settings! Love this!!

db/Cargo.toml
```diff
@@ -7,6 +7,9 @@ edition = "2021"
 
 [dependencies]
 color-eyre = { workspace = true }
+create = "0.1.0"
+migrate = "0.2.0"
+page = "0.0.0"
```
coreyja: I think these might be things you accidentally added with `cargo add` maybe?

I think we can remove em!

db/migrations/20240523181636_userTable.sql
coreyja: Bump on this! I think we want the trailing new line back here

db/migrations/20240531130301_Pages.sql
```diff
@@ -0,0 +1,10 @@
+-- Ensure the UUID extension is enabled
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
```
coreyja: Since this was in the first migrations we can probably leave it off now!

cli/src/add_url.rs
```diff
@@ -0,0 +1,31 @@
...
+    allow_existing: bool,
+) -> Result<AddUrlOutcome> {
+    check_auth_status(pool).await?;
+
+    let result = db::add_url(pool, url, user_id, &allow_existing).await;
+    match result {
+        Ok(page) if page.url.contains("re-adding is allowed") => {
+            Ok(AddUrlOutcome::Existing(page.url.to_string()))
+        }
+        Ok(page) => Ok(AddUrlOutcome::Created(page.url.to_string())),
```
coreyja: Ahh! Nice adding this `AddUrlOutcome` enum in!

Would it be possible to move this enum on lever lower and have it be returned by `db::add_url`?
Cause this bit here where we have to check the string to see if its existing or created is the bit I was trying to avoid really lol!

Since the code in `db::add_url` already 'knows' if its a create or existing it should be able to return the specific answer there, and then these other places can use that return type instead of parsing the strings like this!

Does that make sense?

coreyja (changes requested):
This is looking great! A few nit picky things and then one more change around where we return the 'Created/Existing' enum result!

But looking really good!

cli/src/add_url.rs
Seif-Mamdouh: Makes sense

db/migrations/20240531130301_Pages.sql
Seif-Mamdouh: Do u want me to delete it?

db/src/lib.rs
```diff
@@ -17,6 +17,13 @@ pub struct Page {
     pub url: String,
 }
 
+#[allow(clippy::module_name_repetitions)]
+#[derive(Debug)]
+pub enum AddUrlOutcome {
+    Created(Page),
+    Existing(Page),
+}
+
```
Seif-Mamdouh, line 26: Added the AddUrlOutcome Enum here. Let me know if this is correct

db/migrations/20240531130301_Pages.sql
coreyja: Ya we can leave off/delete this `CREATE EXTENSION` line!

Seif-Mamdouh: Awesome! Done

cli/src/add_url.rs
```diff
@@ -12,11 +12,5 @@ pub async fn add_url(
...
 ) -> Result<AddUrlOutcome> {
     check_auth_status(pool).await?;
-
-    let result = db::add_url(pool, url, user_id, &allow_existing).await;
-    match result {
-        Ok(page) if page.url.contains("re-adding is allowed") => Ok(AddUrlOutcome::Existing(page)),
-        Ok(page) => Ok(AddUrlOutcome::Created(page)),
-        Err(e) => Err(e),
-    }
+    db::add_url(pool, url, user_id, &allow_existing).await
```
Seif-Mamdouh, line 15: This really cleaned up this function here and really simplified it!

cli/src/auth.rs
```diff
@@ -21,3 +22,21 @@ pub async fn check_auth_status(pool: &PgPool) -> color_eyre::Result<()> {
...
+}
+
+
+
+pub fn persist_auth_session(user_id: Uuid) -> color_eyre::Result<()> {
+    let mut file = fs::File::create("auth.txt")?;
+    writeln!(file, "{user_id}")?;
+
+    Ok(())
+}
```
Seif-Mamdouh, line 39: added those two `persist_auth_session` and `get_user_id_from_session` functions` in the `aut mod` since they all check the DB

They are being called in the main file

db/src/lib.rs
```diff
@@ -57,13 +57,13 @@ pub async fn add_url(
...
     } else if *allow_existing {
         println!(
             "URL '{}' exists but re-adding is allowed.",
             upsert_result.url
         );
-        Ok(upsert_result)
+        Ok(AddUrlOutcome::Existing(upsert_result))
     } else {
         Err(color_eyre::eyre::eyre!(
             "URL '{}' already exists and re-adding is not allowed.",
```
Seif-Mamdouh, line 69: Moved the conditions from cli to the DB

cli/src/auth.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::path::Path;
+use uuid::Uuid;
+
+pub async fn check_auth_status(pool: &PgPool) -> color_eyre::Result<()> {
+    let user_path = Path::new("auth.txt");
+    if user_path.exists() {
+        let mut file = fs::File::open("auth.txt")?;
+        let mut contents = String::new();
+        file.read_to_string(&mut contents)?;
+        let user_id = Uuid::parse_str(contents.trim())?;
```
coreyja, line 14: Here we could probably just call `get_user_id_from_session()` right?

cli/src/get_user_id_from_session.rs
```diff
@@ -0,0 +1,12 @@
+use color_eyre::Result;
+use std::fs;
+use std::io::Read;
+use uuid::Uuid;
+
+pub fn get_user_id_from_session() -> Result<Uuid> {
```
coreyja, line 6: And this we can delete right?

coreyja (approved):
Looks great!!

Love how that enum cleaned things up at the end!

PR #7, opened 2024-06-03
Refactoring cli and db

Seif-Mamdouh:
### Refactored files in both the CLI and DB

**Structure of CLI**

- `add_url.rs`
- `auth.rs `(includes functions such as `check_auth_status` , `get_user_id_from_session` & `persist_auth_session`
- `display_user.rs`
- `main.rs`
- `sign_up.rs`

**Structure of DB**

- `add_url.rs`
- `lib.rs`
- `users.rs` (includes functions such as `create_user`, `get_user`, `get_username_by_id`


cli/src/add_url.rs
```diff
@@ -1,16 +1,14 @@
...
-
 pub async fn add_url(
     pool: &PgPool,
     url: &str,
     user_id: Uuid,
     allow_existing: bool,
 ) -> Result<AddUrlOutcome> {
     check_auth_status(pool).await?;
-    db::add_url(pool, url, user_id, &allow_existing).await
+    db::add_url::add_url(pool, url, user_id, &allow_existing).await
```
Seif-Mamdouh: I wasn't able call `add_url` directly on line 13

I tried importing it on line 2 as `use db::add_urls::{add_url, add_url::AddUrlOutcome, PgPool};`

and it gave me this error

Screenshot 2024-06-03 at 4 49 56 PM

Let me know if you wanna chat more!

db/src/lib.rs
```diff
@@ -1,101 +1,9 @@
...
-    user_id: Uuid,
-) -> color_eyre::Result<Option<String>> {
-    let record = sqlx::query!("SELECT user_name FROM Users WHERE user_id = $1", user_id)
-        .fetch_optional(pool)
-        .await?;
-
-    Ok(record.map(|r| r.user_name))
-}
+pub mod add_url;
+pub mod users;
```
Seif-Mamdouh, line 6: calling the mod `add_url` & `users `in lib.rs

cli/src/add_url.rs
coreyja: Ahh nice this is great!

I think your problem in that error message is that `add_url` was _also_ the name of this method you are defining!
So it thought you were trying to call this method recursively instead of calling the one in `db`!

So this full pathing works for me!

db/src/lib.rs
coreyja: Maybe the `add_urls` is just `urls` for the module name?

coreyja (changes requested):
Refactor looks nice! One suggestion on changing a module name but not a big deal!

In terms of feedback on the PR, once thing I like to see more of in the PR description is the 'why' behind the changes!
For this PR the why is probably just to clean things up and get things out of the 'main.rs' and 'lib.rs' files, but good to be explicit about it!

In general the "what" of a PR is usually relatively straightforward from looking at the changes. Here its pretty obvious to me what the code structure is by looking at the diff, that adding it to the PR doesn't add a ton of value for me.
BUT explaining _why_ is something that I can't get at all from reading the code! So its a great thing to include in your PR description!

coreyja (dismissed):
Ooops wrong button last time!

Seif-Mamdouh:
Awesome! Thank you for the feedback. I will include the "why" in the next PR descriptions

cli/src/add_url.rs
```diff
@@ -1,5 +1,5 @@
 use color_eyre::Result;
-use db::add_urls::{add_url, AddUrlOutcome, PgPool};
+use db::urls::{add_url, AddUrlOutcome, PgPool};
```
Seif-Mamdouh, line 2: renamed the mod here

cli/src/add_url.rs
```diff
@@ -3,7 +3,7 @@ use db::urls::{add_url, AddUrlOutcome, PgPool};
 use uuid::Uuid;
 
 use crate::check_auth_status;
-pub async fn add_url_to_db(
+pub async fn append_url_to_db(
```
Seif-Mamdouh, line 6: clippy didn't like the `add_url_to_db` cause the first half `add_url` was named after the function. So I changed it to `append_url_to_db`.

Gotta love clippy😂

cli/src/add_url.rs
Seif-Mamdouh: oh that's silly of me 😂

cli/src/add_url.rs
coreyja: That definitely works!

I think something thats more in the 'spirit' of the lint could be to just call `to_db` or something since we are already scoped to 'add_url' so we don't really need to say that part again in the function name!
Or what I might do is (similar to what we just did in CLI) call the module `url` and then this can be `add_to_db` or something!

I think the _idea_ behind the clippy lint is that repeating something in the module name and the function name is duplicative, and usually just leads to overly long and verbose names. So instead of trying to use a synonym we can try and remove some of the duplication instead!

cli/src/add_url.rs
coreyja: There _might_ have also been a warning there about `add_url` in your import  being 'shadowed' by this function but I'm not positive about that...

By 'shadowed' I mean that your import brought `add_url` into scope, but once you define the function you can't refer to the `db::add_url` as just `add_url` anymore because we 'shadowed' it with the local function

'Shadowing' is also the term used when you have multiple `let`s that have the same name, but may be different types!

```rust
let x: i64 = 6;
let x: String = x.to_string();
```

coreyja (approved):
👍

cli/src/add_url.rs
Seif-Mamdouh: That makes sense.

PR #8, opened 2024-06-04
linear-278-creating a snapshot of the current HTML

Seif-Mamdouh:
This PR enables the system to download the raw HTML of the specified URL from the `Pages Table` using the `reqwest` crate, and then add the data to the `Page Snapshot Table`.

Quick rundown of it works:
- User inserts the url
- system fetches the URL by selecting for the `page_id` `user_id` & `url`
- the URL is scanned through using `reqwest` crate 
- the raw HTML is stored in the `Page Snapshot Table`

Couple of thoughts:
- To persist url uniques and I chose to select `page_id` `user_id` with the designated `URL` in order to avoid duplicates. However I'm not quite sure how things will turn out when the user passes --allow-existing more than once. Happy to chat about more.

-  I wrote a new migration to create a `Page Snapshot Table`. However this means I have to drop the DB, lose data and re create it again. Would loveto chat on how to maybe create new tables with out having to drop data every time 😅


cli/src/add_url.rs
```diff
@@ -3,7 +3,7 @@ use db::urls::{add_url, AddUrlOutcome, PgPool};
 use uuid::Uuid;
 
 use crate::check_auth_status;
-pub async fn append_url_to_db(
+pub async fn append_url(
```
Seif-Mamdouh, line 6: changed the name of this function as requested in the previous PR. Let me know if there's an issues 😅

cli/src/main.rs
```diff
@@ -60,7 +62,11 @@ async fn main() -> color_eyre::Result<()> {
             allow_existing,
         } => {
             let user_id = get_user_id_from_session()?;
-            append_url_to_db(&db_pool, &url, user_id, allow_existing).await?;
+            append_url(&db_pool, &url, user_id, allow_existing).await?;
+            // let outcome = append_url(&db_pool, &url, user_id, allow_existing).await?;
+            // if let AddUrlOutcome::Created(page) | AddUrlOutcome::Existing(page) = outcome {
+            //     process_page_snapshot(&db_pool, page.page_id, user_id, &url).await?;
+            // }
```
Seif-Mamdouh: The commented code is used for testing. Feel free to pull the PR and uncomment the code to see the output.

The output of the code looks something like this 😅 
Screenshot 2024-06-05 at 8 52 07 AM

db/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2021"
 
 [dependencies]
 color-eyre = { workspace = true }
+reqwest = "0.12.4"
```
Seif-Mamdouh: reqwest crate for downloading the RAW HTML

db/migrations/20240604184524_PageSnapshot.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE PageSnapshot (
+    page_snapshot_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    page_id UUID REFERENCES Pages (page_id) NOT NULL,
+    raw_html TEXT NOT NULL,
+    fetched_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP
+    -- title TEXT NOT NULL,
+    -- cleaned_html TEXT NOT NULL,
+    -- cleaned_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP
+    -- markdown_id UUID
```
Seif-Mamdouh, line 9: commented out these properties for now because they weren't being used. I'm planning to use them in the next features

db/src/urls.rs
```diff
@@ -16,6 +18,14 @@ pub enum AddUrlOutcome {
...
 }
 
+#[derive(sqlx::FromRow, Debug)]
+pub struct PageSnapShot {
+    pub page_snapshot_id: Uuid,
+    pub page_id: Uuid,
+    pub raw_html: String,
+    pub fetched_at: Option<chrono::DateTime<chrono::Utc>>,
+}
+
```
Seif-Mamdouh, line 28: PageSnapShot Struct to output the data in `store_raw_html_in_page_snapshot`

db/Cargo.toml
coreyja: Ahh! So I think this is why `cargo deny` is blowing up for you, its doing its job for use here lol 🎉 😆 

One of the big reasons I use `cargo deny` is to prevent pulling in `OpenSSL`. Which is a C library that has been used basically everywhere in the last to do `ssl` which powers `https`

But its a C library which can make compiling the Rust harder, and there is a Rust only alternative that I like to use! And even better `reqwest` can work with either SSL package! So we just need to tell `reqwest` that we want to use different features

https://doc.rust-lang.org/cargo/reference/features.html

Here is where I do reqwest in another project. I think we can leave off the `json` here since we aren't parsing json here!
https://github.com/coreyja/coreyja.com/blob/main/Cargo.toml#L28C1-L31C30

Basically I think we want the `default-features-false` and then the `features` list with `rustls-tls`!

cli/src/main.rs
coreyja: We could build a CLI subcommand that takes in a URL and returns the most recent snapshot HTML if it would be helpful in testing!

But this is also pretty fine with me for now lol

db/migrations/20240604184524_PageSnapshot.sql
coreyja: Nice! Fine to leave them here for reference if you want

But when we go to actually add these we'll want to do it in new migrations instead of editing older ones.

For migrations once they land on the 'main' branch they really should be 'frozen' and never changed! This helps avoid issues like you've seen where a previous migration is edited and SQLx goes crazy until you delete your DB and create it again

So when we want to add these columns later we will do that with NEW migrations and then some `ALTER TABLE` sql commands to add columns. And we'll have to chat about what to do about columns we want to add that are not nullable, but lets cover that when we get there 😆

cli/src/main.rs
coreyja: And actually looking at this code a bit more, I think we just want to add these lines un-commented!

Cause these are what hook up the downloading to the add_url cli command, which is exactly what we want for this ticket!

_if_ we wanted to comment something out we could comment out the println! that prints this all out, just cause it clutters the display. Buttt 🤷 this CLI is really just for us right now, and I think its nicer to see the HTML result as we keep going!

cli/src/main.rs
```diff
@@ -1,13 +1,15 @@
...
-use add_url::append_url_to_db;
+use add_url::append_url;
 
 mod auth;
 use auth::check_auth_status;
 use auth::get_user_id_from_session;
 
 mod sign_up;
+// use db::urls::process_page_snapshot;
+// use db::urls::AddUrlOutcome;
```
coreyja: Nit: Lets remove these!

Don't like leaving commented code on the main branch basically lol

cli/src/main.rs
```diff
@@ -60,7 +62,11 @@ async fn main() -> color_eyre::Result<()> {
             allow_existing,
         } => {
             let user_id = get_user_id_from_session()?;
-            append_url_to_db(&db_pool, &url, user_id, allow_existing).await?;
+            append_url(&db_pool, &url, user_id, allow_existing).await?;
+            // let outcome = append_url(&db_pool, &url, user_id, allow_existing).await?;
+            // if let AddUrlOutcome::Created(page) | AddUrlOutcome::Existing(page) = outcome {
+            //     process_page_snapshot(&db_pool, page.page_id, user_id, &url).await?;
```
coreyja: The `page` within `outcome` is the 'full' Page row right?

Like it has a `page.url` that we could use right? It might be nicer to pass the full `page` down to `process_page_snapshot` so it doesn't need to fetch the URL by id right away!

cli/src/main.rs
```diff
@@ -60,7 +62,11 @@ async fn main() -> color_eyre::Result<()> {
             allow_existing,
         } => {
             let user_id = get_user_id_from_session()?;
-            append_url_to_db(&db_pool, &url, user_id, allow_existing).await?;
+            append_url(&db_pool, &url, user_id, allow_existing).await?;
+            // let outcome = append_url(&db_pool, &url, user_id, allow_existing).await?;
+            // if let AddUrlOutcome::Created(page) | AddUrlOutcome::Existing(page) = outcome {
```
coreyja: Try something for me but it might not work 😆 

I want to see if this works without the `if` so just starting with `let AddUrlOutcome::Created(page) | AddUrlOutcome::Existing(page)`

The reason is a bit subtle lol! What I _want_ to happen is for Rust to error out if we ever added a third type to this enum

What you have now says "If the outcome is a Created or Existing, take its page out and do stuff with it, but if it does NOT match we simply ignore it"

This works great today, cause the enum HAS to be Created or Existing. But what if we added a third variant? In the `if let` version Rust would happily compile and just skip this code path for the third variant.

But I don't think thats _really_ what we want here! We'd RATHER if the Rust compiler errored out when we add the third variant, and tell us we need to update this code. I believe that is what the `let` alone will do! It should work now when we handle both the variants, but if a third is added that we don't handle I believe it will start erroring out!

Does that make sense? And does this code work with just the `let`? lol

coreyja:
> I wrote a new migration to create a Page Snapshot Table. However this means I have to drop the DB, lose data and re create it again. Would loveto chat on how to maybe create new tables with out having to drop data every time 😅

You should NOT need to drop and recreate your DB each time we add a migration 🤞  Thats actually part of the coolness of the migration system!

The reason we've had to before is when we are editing an old migration, and get the sqlx "Migration already run but hash changed" error message

But to just run a new migration you should be able to just do `cargo sqlx migrate run` (or run the `./scripts/migrate.sh` script!

That will just apply the new migration and keep the existing DB!

BUTT you might also run into cases where you need to test a migration, and then realize you need to edit it in your PR.
In that case we probably want to look into 'Reversible Migrations' which help solve that issue!
https://docs.rs/crate/sqlx-cli/latest [Then search for `Reversible Migrations` cant link it 🤦 ]

Basically instead of just creating the migration file, we create an `up` file which is where the migration goes and a `down` file that is used to UNDO the migration.
So for this migration the `down` is likely a `DROP TABLE` for the table we are adding

If you define the down you can run `cargo sqlx migrate revert` and it will 'undo' the last migration! Then you can edit the 'up' and re-run it until you are happy with the results!

Does that make sense and hopefully make this all a bit easier?

coreyja (changes requested):
Looking really good!

Some comments about the migrations and the reqwest Feature flags, but nothing major at all!

cli/src/main.rs
Seif-Mamdouh: Awesome! I will un comment it so we can use them!

db/Cargo.toml
Seif-Mamdouh: Interesting stuff! It's crazy how we can tell Rust which one to use and not to use

db/migrations/20240604184524_PageSnapshot.sql
Seif-Mamdouh: Okay awesome! I will reach out when I get started on this! Thank you for pointing it out for the future!

cli/src/main.rs
Seif-Mamdouh: Okay I'm not sure if I did this entirely correct but I think this maybe could solve the problem?

The reason why is that we maybe use the pattern matching is that because the Rust compiler will enforce handling all of the possible variants and adding a new variant will cause compile time error because it's not handled.

```
let page = match outcome {
                AddUrlOutcome::Created(page) | AddUrlOutcome::Existing(page) => page,
};

process_page_snapshot(&db_pool, page).await?;
```

Let me know what you think!

coreyja: Yes this is PERFECT!! Leaning into the Rust pattern matching is awesome!

Here are a few different versions of this just to compare! There are all just 'syntactic sugar' meaning they will literally be compiled to the same things at the end of the day, just whichever you prefer the style of!

Here is what you wrote (with my syntax highlighting lol) just to have it as a comparison!

image

Here is the first change! Since we only have a single branch on our `match` we can actually just remove it and do the pattern matching in the `let`!

image

And if we wanted to push that one farther we could remove the `outcome` var and do this all on one line!

image

And that last one is _basically_ what you had before without the leading `if` I believe!

One way to think about the `if` at the beginning is `if this pattern matches`, where if the pattern does NOT match it will just just skip that block.
Removing the `if` changes that to a compile error if we know we can't match the pattern!

db/src/urls.rs
```diff
@@ -52,3 +62,48 @@ pub async fn add_url(
...
+        "SELECT page_id, user_id, url FROM Pages WHERE page_id = $1",
+        page_id
+    )
+    .fetch_one(pool)
+    .await?;
+
+    Ok(result)
+}
+
+async fn download_raw_html(_pool: &PgPool, url: &str) -> color_eyre::Result<String> {
```
coreyja: If we don't need the `pool` here we can probably skip passing it in!

coreyja: And actually that makes me wonder where this method should live... Only cause its not _really_ related to the DB...

But I also dont think it makes sense in the CLI so this is great for now at least lol

db/src/urls.rs
```diff
@@ -52,3 +62,48 @@ pub async fn add_url(
         ))
     }
 }
+
+pub async fn fetch_url_from_pages_table(pool: &PgPool, page_id: Uuid) -> color_eyre::Result<Page> {
```
coreyja: Ahh this function is un-used now right? Might just remove it in that case, I think its simple enough to write again if we need it later!

db/src/urls.rs
```diff
@@ -52,3 +62,48 @@ pub async fn add_url(
...
+    pool: &PgPool,
+    page: Page,
+) -> color_eyre::Result<PageSnapShot> {
+    let raw_html = download_raw_html(pool, &page.url).await?;
+    let current_time = chrono::Utc::now();
+
+    let result = sqlx::query_as!(
+        PageSnapShot,
+        "INSERT INTO PageSnapShot (raw_html, fetched_at, page_id) 
+         VALUES ($1, $2, $3) RETURNING page_snapshot_id, page_id, raw_html, fetched_at",
```
coreyja: If you change this to a `RETURNING *` it will fail to compile if we add a new column and do NOT add it to the `PageSnapShot` struct here

Which _can_ be nice if we want that struct to always be the 'full' DB row. No strong feels here just chatting!

db/src/urls.rs
```diff
@@ -52,3 +62,48 @@ pub async fn add_url(
...
+        current_time,
+        page.page_id
+    )
+    .fetch_one(pool)
+    .await?;
+
+    Ok(result)
+}
+
+pub async fn process_page_snapshot(pool: &PgPool, page: Page) -> color_eyre::Result<()> {
```
coreyja, line 93: Love this method since we are going to do more processing after the download :tada:

coreyja (approved):
This looks great!!

I left some cleanup comments but thats all!

I also TURNED OFF the thing that removes reviews when you push every new commit 😆 
That way if you want to push cleanup commits after approval you don't need to either wait for re-approval or bundle it in the next pr lol

db/src/urls.rs
Seif-Mamdouh: Make sense!

PR #9, opened 2024-06-07
linear-279 cleaning up HTML

Seif-Mamdouh:
The PR allows the system to take the raw HTML and clean it up!

to achieve that we used the `readability` crate
docs for `readability`: https://docs.rs/readability/0.3.0/readability/index.html

Raw_HTML before using `readability` would go from something like this:
Screenshot 2024-06-07 at 6 16 40 PM

and clean it up Into this:
Screenshot 2024-06-07 at 6 17 45 PM

update to the migrations files:
- we added `20240607170943_add_cleaned_html_to_pagesnapshot.down.sql` & `20240607170943_add_cleaned_html_to_pagesnapshot.up.sql` to make it flexible to add or remove columns in migrations

### DIFF
--- a/db/src/urls.rs
+++ b/db/src/urls.rs
@@ -68,15 +68,14 @@ pub async fn add_url(
 }
 
 async fn download_raw_html(url: &str) -> color_eyre::Result<String> {
-    let response = reqwest::get(url).await.map_err(color_eyre::Report::from)?;
-    let html = response.text().await.map_err(color_eyre::Report::from)?;
+    let response = reqwest::get(url).await?;
+    let html = response.text().await?;
     Ok(html)
 }
 
 fn clean_raw_html(raw_html: &str, url: &Url) -> color_eyre::Result<String> {
     let mut raw_html_cursor = std::io::Cursor::new(raw_html);
-    let article = extractor::extract(&mut raw_html_cursor, url)
-        .map_err(|e| color_eyre::eyre::eyre!(e.to_string()))?;
+    let article = extractor::extract(&mut raw_html_cursor, url)?;
     Ok(article.content)
 }
 
--- a/db/src/urls.rs
+++ b/db/src/urls.rs
@@ -87,7 +87,7 @@ async fn store_raw_html_in_page_snapshot(
     let raw_html = download_raw_html(&page.url).await?;
     let current_time = chrono::Utc::now();
     let url = Url::parse(&page.url)?;
-    let cleaned_html = clean_raw_html(&raw_html, &url).await?;
+    let cleaned_html = clean_raw_html(&raw_html, &url)?;
 
     let result = sqlx::query_as!(
         PageSnapShot,
--- a/db/src/urls.rs
+++ b/db/src/urls.rs
@@ -1,10 +1,10 @@
-use reqwest;
 use readability;
 use readability::extractor;
-use url::Url;
+use reqwest;
 pub use sqlx;
 use sqlx::types::chrono;
 pub use sqlx::PgPool;
+use url::Url;
 use uuid::Uuid;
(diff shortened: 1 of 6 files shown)


db/src/urls.rs
```diff
@@ -69,19 +73,29 @@ async fn download_raw_html(url: &str) -> color_eyre::Result<String> {
     Ok(html)
 }
 
+fn clean_raw_html(raw_html: &str, url: &Url) -> color_eyre::Result<String> {
+    let mut raw_html_cursor = std::io::Cursor::new(raw_html);
+    let article = extractor::extract(&mut raw_html_cursor, url)
+        .map_err(|e| color_eyre::eyre::eyre!(e.to_string()))?;
+    Ok(article.content)
+}
+
```
Seif-Mamdouh, line 81: Used `readability::extractor` to extract the HTML content from the URL

if you want to checkout the docs: https://docs.rs/readability/0.3.0/readability/extractor/struct.Product.html

db/src/urls.rs
```diff
@@ -69,19 +73,29 @@ async fn download_raw_html(url: &str) -> color_eyre::Result<String> {
...
 async fn store_raw_html_in_page_snapshot(
     pool: &PgPool,
     page: Page,
 ) -> color_eyre::Result<PageSnapShot> {
     let raw_html = download_raw_html(&page.url).await?;
     let current_time = chrono::Utc::now();
+    let url = Url::parse(&page.url)?;
```
Seif-Mamdouh, line 88: used the `url` crate to parse the URL String into a URL a object. This is necessary because the `clean_raw_html` requires a url object as one of the function arguments.

By parsing the URL string into a URL string we ensure the URL that the user inserted is valid.

docs: https://docs.rs/url/latest/url/

db/src/urls.rs
coreyja: It's interesting that this extract takes the URL in, but on second thought it makes sense!

It's probably used at some point to (at a miniumum) convert relative links to absolute ones, or things like that.
I'm actually running into a similar (but reverse problem kinda) where in my blog I'm going from Markdown to HTML and currently when I link to an image its usually with a relative link like `../images/something.png` and when the browser sees that it knows to use the current url as the 'base'
My problem was that I wanted to add a CDN to the middle, and the CDN needs the full url to the image. And of course in my markdown -> html code path I don't currently have the url! So my next refactor there is going to be piping the URL down farther lol

Just rambling! No changes here, looks great!

db/src/urls.rs
```diff
@@ -69,19 +73,29 @@ async fn download_raw_html(url: &str) -> color_eyre::Result<String> {
     Ok(html)
 }
 
+fn clean_raw_html(raw_html: &str, url: &Url) -> color_eyre::Result<String> {
+    let mut raw_html_cursor = std::io::Cursor::new(raw_html);
+    let article = extractor::extract(&mut raw_html_cursor, url)
+        .map_err(|e| color_eyre::eyre::eyre!(e.to_string()))?;
```
coreyja: Do we need the `map_err` here? I think some of this is what the `?` operator with color_eyre does for us so we might not need it 🤞

coreyja (approved):
This is perfect! Love it!

Left one comment about maybe being able to cleanup the error handling, but not a big deal if it doesn't work/compile lol

db/src/urls.rs
Seif-Mamdouh: Your next refactor on coreyja.com?

I can take a look if you want me to?

db/src/urls.rs
Seif-Mamdouh: you are correct. I removed it

PR #10, opened 2024-06-10
adding clean_html to MD file

Seif-Mamdouh:
This PR allows the system to take `cleaned_html` from `PageSnapShot Table` and created a Markdown version of the `cleaned_html`.

We also created a migration from Markdown table. 
We have `db/migrations/20240610191311_Markdown.up.sql` and `db/migrations/20240610191311_Markdown.down.sql`
the reason why is because we want to make it flexible to add, remove or update migrations.

### DIFF
--- a/db/src/urls.rs
+++ b/db/src/urls.rs
@@ -86,6 +86,27 @@ fn clean_raw_html(raw_html: &str, url: &Url) -> color_eyre::Result<String> {
     Ok(article.content)
 }
 
+async fn store_markdown(
+    pool: &PgPool,
+    page_snapshot_id: Uuid,
+    cleaned_html: &str,
+) -> color_eyre::Result<Markdown> {
+    let markdown_content = html2md::parse_html(cleaned_html);
+    println!("Markdown content: {markdown_content:?}");
+
+    let markdown_result = sqlx::query_as!(
+        Markdown,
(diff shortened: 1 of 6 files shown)


db/src/urls.rs
```diff
@@ -79,6 +86,27 @@ fn clean_raw_html(raw_html: &str, url: &Url) -> color_eyre::Result<String> {
     Ok(article.content)
 }
 
+async fn store_markdown(
+    pool: &PgPool,
+    page_snapshot_id: Uuid,
+    cleaned_html: &str,
+) -> color_eyre::Result<Markdown> {
+    let markdown_content = html2md::parse_html(cleaned_html);
```
Seif-Mamdouh, line 94: Used `parse_html` to parse all the content from the article 

docs: https://docs.rs/html2md/latest/html2md/fn.parse_html.html

db/src/urls.rs
```diff
@@ -100,6 +128,9 @@ async fn store_raw_html_in_page_snapshot(
     .fetch_one(pool)
     .await?;
 
+    let markdown_result = store_markdown(pool, result.page_snapshot_id, &cleaned_html).await?;
+    println!("Markdown result: {markdown_result:?}");
+
```
Seif-Mamdouh, line 133: printing out the result for testing purposes

coreyja (approved):
Love it, this looks great! No comments at all 🎉

PR #11, opened 2024-06-11
Add `sqlx prepare --check` to CI

coreyja:
This will help prevent forgetting to run `sqlx prepare` before committing changes to the database schema, or adding new queries in Rust.

This way we will always have up to do `.sqlx` files in the repository, that that it can be built without needing to connect to the database.


Seif-Mamdouh (approved):
Awesome! This is great because I forgot to run `cargo sqlx prepare --workspace` yesterday and clippy was giving me trouble. But this is very helpful!

PR #12, opened 2024-06-11
Take the MD file and summarize the content using AI

Seif-Mamdouh:
In this PR the system takes the MD content from the `Mardown table` and summarizes it using the power of AI!

To achieve this we used openai crate 
documentation: https://docs.rs/openai/1.0.0-alpha.14/openai/chat/struct.ChatCompletionBuilder.html

Screenshot 2024-06-11 at 3 46 53 PM

### DIFF
--- a/db/Cargo.toml
+++ b/db/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2021"
 [dependencies]
 color-eyre = { workspace = true }
 html2md = "0.2.14"
+openai = { version = "1.0.0-alpha.14", features = ["rustls"], default-features = false }
 readability = { version = "0.3.0", default-features = false }
 reqwest = { version = "0.12.4", features = [
   "rustls-tls"
--- a/db/src/openai_utils.rs
+++ b/db/src/openai_utils.rs
@@ -1,0 +1,38 @@
+use color_eyre::Result;
+use openai::chat::{ChatCompletionBuilder, ChatCompletionMessage, ChatCompletionMessageRole};
+use std::env;
+
+pub async fn generate_summary(content: &str) -> Result<String> {
+    let api_key = env::var("OPEN_AI_API_KEY").expect("OPEN_AI_API_KEY must be set");
+    openai::set_key(api_key);
+
+    let messages = vec![
+        ChatCompletionMessage {
+            role: ChatCompletionMessageRole::System,
+            content: Some("You are a helpful assistant.".to_string()),
+            name: None,
+            function_call: None,
+        },
+        ChatCompletionMessage {
+            role: ChatCompletionMessageRole::User,
+            content: Some(format!("Summarize the following article and make sure to highlight the important parts: {}", content)),
+            name: None,
+            function_call: None,
+        },
+    ];
+    let request = ChatCompletionBuilder::default()
+        .model("gpt-4".to_string())
+        .messages(messages)
+        .max_tokens(4096u64)
+        .temperature(0.7)
+        .top_p(1.0)
+        .build()?;
+
+    let response = openai::chat::ChatCompletion::create(&request).await?;
+    let summary = response.choices[0]
+        .message
+        .content
+        .clone()
+        .unwrap_or_default();
+    Ok(summary)
+}
--- a/db/src/openai_utils.rs
+++ b/db/src/openai_utils.rs
@@ -8,14 +8,8 @@ pub async fn generate_summary(content: &str) -> Result<String> {
 
     let messages = vec![
         ChatCompletionMessage {
-            role: ChatCompletionMessageRole::System,
-            content: Some("You are a helpful assistant.".to_string()),
-            name: None,
-            function_call: None,
-        },
-        ChatCompletionMessage {
             role: ChatCompletionMessageRole::User,
-            content: Some(format!("Summarize the following article and make sure to highlight the important parts: {content}")),
+            content: Some(format!("Provide a concise summary of the following article that can be read and understood in 2 minutes, focusing on the main points and essential details: {content}")),
             name: None,
             function_call: None,
         },
@@ -25,7 +19,6 @@ pub async fn generate_summary(content: &str) -> Result<String> {
         .messages(messages)
         .max_tokens(4096u64)
         .temperature(0.7)
-        .top_p(1.0)
         .build()?;
 
     let response = openai::chat::ChatCompletion::create(&request).await?;
--- a/db/src/openai_utils.rs
+++ b/db/src/openai_utils.rs
@@ -21,7 +21,7 @@ pub async fn generate_summary(content: &str) -> Result<String> {
         },
     ];
     let request = ChatCompletionBuilder::default()
-        .model("gpt-4".to_string())
+        .model("gpt-4o".to_string())
         .messages(messages)
         .max_tokens(4096u64)
         .temperature(0.7)
--- a/db/src/openai_utils.rs
+++ b/db/src/openai_utils.rs
@@ -15,7 +15,7 @@ pub async fn generate_summary(content: &str) -> Result<String> {
         },
         ChatCompletionMessage {
             role: ChatCompletionMessageRole::User,
-            content: Some(format!("Summarize the following article and make sure to highlight the important parts: {}", content)),
+            content: Some(format!("Summarize the following article and make sure to highlight the important parts: {content}")),
             name: None,
             function_call: None,
         },
--- a/db/src/urls.rs
+++ b/db/src/urls.rs
@@ -4,8 +4,16 @@ use reqwest;
 pub use sqlx;
 use sqlx::types::chrono;
 pub use sqlx::PgPool;
+use std::collections::HashMap;
+use std::env;
 use url::Url;
 use uuid::Uuid;
+
+use openai::chat::{
+    ChatCompletionBuilder, ChatCompletionMessage, ChatCompletionMessageRole, ChatCompletionRequest,
+};
+use openai::completions::Completion;
+use openai::completions::CompletionRequest;
 
 #[derive(sqlx::FromRow, Debug)]
 pub struct Page {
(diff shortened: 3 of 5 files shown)


db/Cargo.toml
```diff
@@ -8,6 +8,7 @@ edition = "2021"
 [dependencies]
 color-eyre = { workspace = true }
 html2md = "0.2.14"
+openai = { version = "1.0.0-alpha.14", features = ["rustls"], default-features = false }
```
Seif-Mamdouh, line 11: we are using feature `rustls` to make sure we are not pulling in openssl

db/src/urls.rs
```diff
@@ -83,7 +85,7 @@ async fn download_raw_html(url: &str) -> color_eyre::Result<String> {
 fn clean_raw_html(raw_html: &str, url: &Url) -> color_eyre::Result<String> {
     let mut raw_html_cursor = std::io::Cursor::new(raw_html);
     let article = extractor::extract(&mut raw_html_cursor, url)?;
-    Ok(article.content)
+    Ok(article.text)
```
Seif-Mamdouh: in extractor function from readability crate

I'm returning `article.text` because `.text` contains the plain text content of the article, with all HTML tags removed. It is useful because I have seen the AI respond better to pure plain text.

In the beginning we had it as `article.content`. `content` contains the cleaned HTML content of the article. It retains the HTML structure but removes unnecessary elements like ads, sidebars. However it can giet very messy if the article has pictures that model can't read through it yet. 

I didn't find the description of what `text` & `content` really does. I found that one by trying things out.
docs: https://docs.rs/readability/latest/readability/extractor/struct.Product.html

Maybe we can play around the AI and make it scan through images (if it sees one) include it the summary 🤔.

However for now I chose to use `.text` to help the AI output better responses! 

Happy to chat more to see your input!

db/src/openai_utils.rs
```diff
@@ -0,0 +1,38 @@
...
+        .build()?;
+
+    let response = openai::chat::ChatCompletion::create(&request).await?;
+    let summary = response.choices[0]
+        .message
+        .content
+        .clone()
+        .unwrap_or_default();
+    Ok(summary)
+}
```
Seif-Mamdouh, line 31: We used `chat` from openai crate.

Utilized the struct ChatCompletionMessage to create a message for the GPT. In the struct we have role and content.

`ChatCompletionMessageRole::System` to identify the app's role
`ChatCompletionMessageRole::User` to identify the user's role 

docs: https://docs.rs/openai/1.0.0-alpha.14/openai/chat/struct.ChatCompletionMessage.html

The `ChatCompletionBuilder` constructs the request with these messages.
docs: https://docs.rs/openai/1.0.0-alpha.14/openai/chat/struct.ChatCompletionBuilder.html

`ChatCompletion::create` method to send a request for OpenAI API to generate the `ChatCompletion` based on the messages that is provided from `ChatCompletionMessage`
docs: https://docs.rs/openai/1.0.0-alpha.14/openai/chat/type.ChatCompletion.html#method.create

db/src/openai_utils.rs
```diff
@@ -0,0 +1,38 @@
...
+            name: None,
+            function_call: None,
+        },
+    ];
+    let request = ChatCompletionBuilder::default()
+        .model("gpt-4o".to_string())
+        .messages(messages)
+        .max_tokens(4096u64)
+        .temperature(0.7)
+        .top_p(1.0)
```
coreyja: The `max_tokens` and `temperature` I'm familiar with, what is `top_p`/what does it control?

db/src/urls.rs
coreyja: Ya love this!!

And we had pretty similar thoughts!

I was thinking that we could use the Markdown that we convert after this as the 'text version'!
My thinking is that Markdown is much closer to plain text, but can keep just a _bit_ of the formatting like titles and stuff so that the LLM can use those to help it ou!

Like I think knowing there is a heading call 'Migrations' (or whatever) is good to know over the work just being in the middle of a paragraph lol

Does that make sense? What do you think about using the markdown for that?

BUT I'm ok with using the plain text too, if you think it will be better!
But I do think I'd like to at least _store_ the cleaned html, so if we want to save the text too how about we do it in another column?

db/src/openai_utils.rs
```diff
@@ -0,0 +1,38 @@
...
+    let api_key = env::var("OPEN_AI_API_KEY").expect("OPEN_AI_API_KEY must be set");
+    openai::set_key(api_key);
+
+    let messages = vec![
+        ChatCompletionMessage {
+            role: ChatCompletionMessageRole::System,
+            content: Some("You are a helpful assistant.".to_string()),
+            name: None,
+            function_call: None,
+        },
```
coreyja: Would be curious to try removing this and seeing what the results are!

Not cause I think its bad, I'm just interesting in finding out how helpful it is lol

Thats another thing to keep in mind :thinking:

How would be go about evaluating this "prompt" in the future so that we can tune it? Doesn't have to be something we can do/answer today, but would be really cool to have a way to like test out a prompt on a bunch of articles we've already saved or something 🤔 
Probably something for when we get to the web side!

db/src/urls.rs
Seif-Mamdouh: Your suggestion is to create a `.txt` file from the `Markdown file`? I just want to make sure that I'm on the same page as you.

db/src/openai_utils.rs
Seif-Mamdouh: The .top_p parameter in the context of the OpenAI API is used to control nucleus sampling, which is a technique for generating text. 

reference: https://community.openai.com/t/temperature-top-p-and-top-k-for-chatbot-responses/295542

It's basically letting the model consider all smallest amount of possible tokens, the model considers only the smallest set of tokens whose cumulative probability is greater than or equal to the top_p value

in my code I have it `top_p(1.0)` which means the model will consider all possible tokens which doesn't really make sense to have it, so I might remove it 😃

db/src/openai_utils.rs
Seif-Mamdouh: I removed the `ChatCompletionMessageRole::System`, the results are pretty much the same!

When it comes to the prompt, I agree this prompt can get def be better. I just did this for now just to test if it works but I can spend some time today working on the prompt to ensure that we can get better results for the user.

Let me know if you have any recommendations for prompt engineering

db/src/urls.rs
coreyja: No sorry!

Two suggestions take your pick between them!

- Don't store the text and pass the markdown to the LLM
- Store the txt in a new column and pass that to the LLM

Originally I was thinking we'd use the markdown, and was kinda why we did that step before this llm one!

But if you think the txt leads to better summaries I'm down to do that! Just don't want to _lose_ the cleaned html if that makes sense!

db/src/openai_utils.rs
coreyja: Not _really_ lol

I do have some custom prompts for ChatGPT which I like so I'll share!

> Give short and informative answers. No need for fluff

So maybe something like that would be helpful to add?

db/src/urls.rs
Seif-Mamdouh: We are already passing down the markdown to LLM and the results are looking good in my opinion. So I will go with option for now.

Worst case we can circle back and make the change and see where things go.

For MVP goals we are fine for now. I think 🧐

db/src/openai_utils.rs
```diff
@@ -7,15 +7,9 @@ pub async fn generate_summary(content: &str) -> Result<String> {
...
-            role: ChatCompletionMessageRole::System,
-            content: Some("You are a helpful assistant.".to_string()),
-            name: None,
-            function_call: None,
-        },
         ChatCompletionMessage {
             role: ChatCompletionMessageRole::User,
-            content: Some(format!("Summarize the following article and make sure to highlight the important parts: {content}")),
+            content: Some(format!("Provide a concise summary of the following article that can be read and understood in 2 minutes, focusing on the main points and essential details: {content}")),
             name: None,
```
Seif-Mamdouh, line 13: Let me know what you think about this prompt. I like better because it's a lot more short and concise. 

The User case I imagine is that someone has 5 minutes to read an article and talk to their friend about it, they don't need large paragraphs summary. They just need key highlighted bullet points.

this is an example output for this article: https://daily.dev/blog/multi-tenant-database-design-patterns-2024
Screenshot 2024-06-12 at 12 22 37 PM

db/src/urls.rs
coreyja: Can this go back to doing `article.content` then? So that we still save the cleaned html as opposed to just the text?

I don't want to 'lose' the HTML cause I think it will be useful later if we want to be smarter and parse things out (like the title tag!)

coreyja: > We are already passing down the markdown to LLM and the results are looking good in my opinion.

Ohhh! 🤦  I looked at the commits from this AM again and I def just read them wrong 😆 

I didn't realize the markdown was being passed to the LLM, I thought _just_ this `text` was. Sorry!

But is the cleaned html + markdown not giving as good results as the `text + markdown`?

I'd be interested in potentially also just feeding in the raw 'text' at that point... Not sure if going from text -> markdown does much.. Actually does it do _anything_ when we pass the `article.text` to the markdown converter 😆 ? Is it basically the same thing back out lol?

Seif-Mamdouh: Yes we can go back to `article.content`. 

Just to make sure I'm explaining this correctly if we return `article.text` in the in the `clean_raw_html` function, the MD file contains just the pure texts. if we use `article.content` we return a cleaned up HTML in the MD file. That includes all the `HTML tags`.

So I ran a couple of experiments.  if we want to keep the "_cleaned html_" into the MD file then we can keep it. My thesis is the AI when it scans "cleaned_html" it returns a longer response. 

If our objective is summarizing and keeping it concise then creating a `text column` in `MD table` is a good idea.

I think to figure out a the best approach, we should hop on a quick 5 min slack huddle so I can demo it and we can together take a good decision!

coreyja: Ya I think a huddle probably makes sense!

I'm going into the city today, so I did a calendar invite just to plan a bit better! Let me know if 1pm doesn't work for you though and happy to move it around!

db/src/openai_utils.rs
coreyja: Love this prompt!


And definitely agree on the use case you outlined! 💯

db/src/urls.rs
```diff
@@ -86,7 +86,7 @@ async fn download_raw_html(url: &str) -> color_eyre::Result<String> {
 fn clean_raw_html(raw_html: &str, url: &Url) -> color_eyre::Result<String> {
     let mut raw_html_cursor = std::io::Cursor::new(raw_html);
     let article = extractor::extract(&mut raw_html_cursor, url)?;
-    Ok(article.text)
+    Ok(article.content)
```
Seif-Mamdouh, line 89: great call. I think we made a good decision.

coreyja (approved):
Love this! 💯 Nicely done!

PR #13, opened 2024-06-13
Ticket 283 category create tags

Seif-Mamdouh:
This PR we created a second prompt for the AI to generate a category for the article that user wanted to summarize.

The AI generated category is later inserted in the DB under `Category Table`.

### **Quick small updates**

- Created a new migration table named `Category`
- As discussed in our quick chat earlier today, in our `Category` table we decided to create a `Foreign Key reference` to the `markdown_id` in the `Markdown` table.
- Screenshot 2024-06-13 at 5 00 21 PM

- we made the decision to change in our `clean_raw_html` function to return a `article.content` instead of `article.text`
- Screenshot 2024-06-13 at 5 02 15 PM

### DIFF
--- a/db/migrations/20240613150913_category.up.sql
+++ b/db/migrations/20240613150913_category.up.sql
@@ -1,5 +1,5 @@
 -- Add migration script here
 CREATE TABLE IF NOT EXISTS Category (
-    note_id UUID PRIMARY KEY,
+    markdown_id UUID PRIMARY KEY REFERENCES Markdown(markdown_id),
     category TEXT
 );
--- a/db/migrations/20240613150913_category.up.sql
+++ b/db/migrations/20240613150913_category.up.sql
@@ -1,0 +1,5 @@
+-- Add migration script here
+CREATE TABLE IF NOT EXISTS Category (
+    note_id UUID PRIMARY KEY,
+    category TEXT
+);
--- a/db/src/openai_utils.rs
+++ b/db/src/openai_utils.rs
@@ -34,7 +34,8 @@ pub async fn generate_summary(content: &str) -> Result<String> {
 }
 
 pub async fn generate_categories(content: &str) -> Result<String> {
-    let prompt = format!("Provide a unique category title for the following article. Do not exceed two words: {content}");
+    let prompt = format!("Provide unique categories for the following article. Do not exceed two words per category. Separate categories with commas:
+: {content}");
     let response = generate_openai_response(&prompt, 100).await?;
     let category = response
         .split(',')
--- a/db/src/openai_utils.rs
+++ b/db/src/openai_utils.rs
@@ -6,14 +6,12 @@ async fn generate_openai_response(prompt: &str, max_tokens: u64) -> Result<Strin
     let api_key = env::var("OPEN_AI_API_KEY").expect("OPEN_AI_API_KEY must be set");
     openai::set_key(api_key);
 
-    let messages = vec![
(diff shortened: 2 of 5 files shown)


db/src/openai_utils.rs
```diff
@@ -2,30 +2,45 @@ use color_eyre::Result;
...
+    let prompt = format!("Provide a unique category title for the following article. Do not exceed two words: {content}");
+    let response = generate_openai_response(&prompt, 100).await?;
+    let category = response
+        .split(',')
+        .next()
+        .unwrap_or_default()
+        .trim()
+        .to_string();
+    Ok(category)
 }
```
Seif-Mamdouh, line 47: cleaned up this function a little bit. 
made a `generate_openai_response` helper function that can be called in `generate_summary` && `generate_categories`

db/migrations/20240613150913_category.up.sql
```diff
@@ -0,0 +1,5 @@
+-- Add migration script here
+CREATE TABLE IF NOT EXISTS Category (
+    markdown_id UUID PRIMARY KEY REFERENCES Markdown(markdown_id),
```
Seif-Mamdouh, line 3: Not sure if I should make a `markdown_id` a `PRIMARY KEY` or not. Let me know what you think

coreyja: Ahh! I think we probably want a `category_id` here to be the primary key, I usually like having a specific id for each table

Also, we might change how we attach categories to `Markdowns` eventually, and I think it will be easier to do that if each has its own id as a primary key!

db/src/openai_utils.rs
```diff
@@ -2,30 +2,45 @@ use color_eyre::Result;
...
+    Ok(content)
+}
+
+pub async fn generate_summary(content: &str) -> Result<String> {
+    let prompt = format!("Provide a concise summary of the following article that can be read and understood in 2 minutes, focusing on the main points and essential details: {content}");
+    generate_openai_response(&prompt, 4096).await
+}
+
+pub async fn generate_categories(content: &str) -> Result<String> {
+    let prompt = format!("Provide a unique category title for the following article. Do not exceed two words: {content}");
```
coreyja: Might be interesting to see if we can get _multiple_ categories out!

Maybe the prompt is something like

```
Provide unique categories for the following article. Do not exceed two words per category. Separate categories with commas:
```

And then you can split on commas after and hopefully get a few categories back out?

db/src/openai_utils.rs
```diff
@@ -2,30 +2,45 @@ use color_eyre::Result;
...
+    let response = generate_openai_response(&prompt, 100).await?;
+    let category = response
+        .split(',')
+        .next()
```
coreyja, line 42: Oh didn't notice this until I made my first comment 😆 

Are there multiple categories coming out of here right now?

Seif-Mamdouh: As if right now no, but I put there just in case

db/migrations/20240613150913_category.up.sql
Seif-Mamdouh: awesome! For the category_id it is the generated UUID inside the `store_markdown`?

for example it will be something like this:

```
async fn store_category(
) -> color_eyre::Result<Category> {
let category_id = //generate UUID
}
```

coreyja (approved):
💯

PR #14, opened 2024-06-17
Ticket 287 make embedding category table

Seif-Mamdouh:
In this PR we use `openai::embeddings` rust crate to generate text embeddings for each category and insert the text embeddings in the `embeddings` column in the `Category` Table

Example Usage:
Screenshot 2024-06-17 at 1 23 25 PM

### DIFF
--- a/db/migrations/20240613150913_category.up.sql
+++ b/db/migrations/20240613150913_category.up.sql
@@ -1,5 +1,6 @@
 -- Add migration script here
 CREATE TABLE IF NOT EXISTS Category (
     markdown_id UUID PRIMARY KEY REFERENCES Markdown(markdown_id),
-    category TEXT
+    category TEXT,
+    embedding DOUBLE PRECISION[]
 );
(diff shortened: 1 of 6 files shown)


db/migrations/20240613150913_category.up.sql
```diff
@@ -0,0 +1,6 @@
+-- Add migration script here
+CREATE TABLE IF NOT EXISTS Category (
+    markdown_id UUID PRIMARY KEY REFERENCES Markdown(markdown_id),
+    category TEXT,
+    embedding DOUBLE PRECISION[]
```
Seif-Mamdouh: added `embedding` column. It stores the array of floats

coreyja: I _LOVE_ that you figured this out on your own nicely done!!

We might eventually use a Postgres plugin and something slightly different here, but a problem for another day!
This is perfect!

coreyja (approved):

PR #15, opened 2024-06-21
generating test use case for display user function

Seif-Mamdouh:
In this PR I wrote a series of async tests to test the functionality of the cli system subcommands in the `cli/main_test.rs`

key highlights functions in `cli/main_test.rs`:

- Added a `run_test_command` to parse the subcommands and establish database connection. This function is used across all other subcommand functions to avoid being DRY!

`run_test_command` is used in 

- `test_display_users_command`: Tests the display-users command.
- `test_sign_up_command`: Tests the signup command with a username argument.
- `test_add_url_command`: Tests the add-url command with URL and --allow-existing arguments.
- `test_status_command`: Tests the status command.

I'm not quite sure what is the best way to write up tests in Rust! However I'm looking forward to the feedback!

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -74,7 +74,6 @@ async fn main() -> color_eyre::Result<()> {
     Ok(())
 }
 
-
 #[cfg(test)]
 mod tests {
     include!("main_test.rs");
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -40,11 +40,6 @@ enum Command {
     },
 }
 
-#[cfg(test)]
-mod tests {
-    include!("main_test.rs");
-}
-
 #[tokio::main]
 async fn main() -> color_eyre::Result<()> {
     let args = KnowledgeArgs::parse();
@@ -77,4 +72,10 @@ async fn main() -> color_eyre::Result<()> {
     }
 
     Ok(())
+}
+
+
+#[cfg(test)]
+mod tests {
+    include!("main_test.rs");
 }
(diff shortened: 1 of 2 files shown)


coreyja:
Hey sorry its taken me so long to get to this PR!

Ya let's definitely talk about testing, I think what you have here is a great start for some CLI tests! And I think I can help fill in some of the missing pieces for ya!

One thing that is missing from these in general is the expectations, which 'prove' that the system behaves like it should when used a certain way. So after the add article command, we should look in the DB and make sure the article was inserted

We can probably do that with some `sqlx` queries and then using `assert_eq!` or `assert!` to make sure things you expect to happen do!
https://doc.rust-lang.org/std/macro.assert_eq.html

I also did start some DB setup code for tests, which gives each tests it's own database so that tests don't run in your normal 'development' database, and so that tests don't interfere with one another!
So you should be able to use this function in your tests to get a DB Pool!
Except not really cause I didn't make it public 😆 
https://github.com/coreyja/knowledge/blob/main/db/src/lib.rs#L57
But we could move that around a bit and use that!

----

Buttt we might also just focus our energy on the Web side now lol and I think we will use PlayWright for that: https://playwright.dev/docs/writing-tests

Which controls a real browser and does actions as close to as a User as is reasonable lol
So we'll like actually sign up for an account as a User in a test and then click around and make sure things happen right!
But I haven't figured out exactly what we will do about the ChatGPT calls in tests like that so I'll need to think about that 🤔  We _could_ run them but then might want to run them on 3.5 so its cheaper lol
But would also be nice to have a way to not make them with not exactly sure how I think we should do that so will think about it!

PR #16, opened 2024-06-22
Web Boilerplate + Auth + Deploy

coreyja:
OKK! This is a bit bug lol, but it stands up a very simple Web Server, and wires up User Auth!


.dockerignore
```diff
@@ -0,0 +1,2 @@
+/target
+.envrc
```
coreyja, line 2: We don't need these things in the Docker conatiner for deploys so add these to the ignore file to make builds faster!

(The target dir gets big do adding it is slow and doesn't help at all, and the envrc file is local secrets we don't want in prod!)

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,13 @@
+on:
+  workflow_call:
```
coreyja, line 2: This Github Action runs when we call it from other actions, nice for making re-usable Github Actions workflows

This is the workflow that will auto-deploy all our changes to Fly.io

.github/workflows/main.yml
```diff
@@ -0,0 +1,17 @@
+name: CI / CD
+
+on:
+  push:
+    branches:
+      - main
```
coreyja, line 6: This is the main branch CI/CD

It runs on pushes to the main branch only

Below you'll see if uses the two other workflows we have and makes sure the tests run BEFORE the deploy starts

.github/workflows/rust_tests.yml
```diff
@@ -1,9 +1,7 @@
 name: Rust Tests
 
 on:
-  push:
-    branches:
-      - main
```
coreyja, line 6: Instead of also running the tests on the main branch, we make this a re-usable workflow and call it from the `main.yml` workflow!

.vscode/settings.json
```diff
@@ -1,8 +1,8 @@
...
-    "editor.formatOnSave": true,
-    "rust-analyzer.rustfmt.enableRangeFormatting": false,
-    "[rust]": {
-        "editor.defaultFormatter": "matklad.rust-analyzer"
-    },
-    "files.insertFinalNewline": true
+  "editor.formatOnSave": true,
+  "rust-analyzer.rustfmt.enableRangeFormatting": false,
+  "[rust]": {
+    "editor.defaultFormatter": "rust-lang.rust-analyzer"
```
coreyja, line 5: Oh I think you might have been using a non-official Rust Analyzer plugin

This one is better!

Cargo.toml
```diff
@@ -9,8 +12,10 @@ tokio = { version = "1", features = ["full"] }
 tracing = "0.1.40"
 url = "2.5.0"
 uuid = { version = "1.8.0", features = ["v4"] }
+chrono = { version = "0.4.23", default-features = false, features = ["serde"] }
 
 [workspace.lints.clippy]
 pedantic = "deny"
 missing_errors_doc = "allow"
 missing_panics_doc = "allow"
+items_after_statements = "allow"
```
coreyja, line 21: A macro we are using for `axum` was tripping this and easiest to just disable the lint lol

Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM lukemathwalker/cargo-chef:latest-rust-1 AS chef
```
coreyja, line 1: This was mostly copied from one of my other projects!

Docker file to deploy our service

We use `cargo-chef` to help with Dockerfile caching and speed up build times a bit!

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+FROM chef AS planner
+COPY . .
+RUN cargo chef prepare --recipe-path recipe.json
+
+FROM chef AS builder 
+RUN rustc --version; cargo --version; rustup --version
+
+# RUN curl -sLO https://github.com/tailwindlabs/tailwindcss/releases/latest/download/tailwindcss-linux-x64 && \
+#   chmod +x tailwindcss-linux-x64 && \
+#   mv tailwindcss-linux-x64 tailwindcss
```
coreyja, line 13: Commented out cause we are taking the easy approach of using Tailwind CSS from a CDN, but we will likely switch this eventually

db/build.rs
```diff
@@ -0,0 +1,5 @@
+// generated by `sqlx migrate build-script`
+fn main() {
+    // trigger recompilation when a new migration is added
+    println!("cargo:rerun-if-changed=migrations");
+}
```
coreyja, line 5: Ahh! This should fix the issue where we would change a mirgation and cargo wouldn't re-compile and realize the change lol

Now it will!

db/migrations/20240619190000_AddJobs.up.sql
```diff
@@ -0,0 +1,13 @@
+-- Add migration script here
+CREATE TABLE
+  Jobs (
```
coreyja, line 3: We did this one together! Just copied from the `cja` stuff basically

This table records one specific piece of work that we want to do

db/migrations/20240619190052_AddCrons.up.sql
```diff
@@ -0,0 +1,17 @@
+-- Add migration script here
+CREATE TABLE
+  Crons (
```
coreyja, line 3: Table to store the cron jobs that run on some schedule

db/migrations/20240622181744_AddSessions.up.sql
```diff
@@ -0,0 +1,9 @@
+-- Add migration script here
+CREATE TABLE
+  Sessions (
```
coreyja, line 3: This is for User Auth.

When you go to a page we will create a Session row for you and add the `session_id` as a private Cookie on your browser.

We will then use this session_id cookie to find your Session. This is how we will know a user is logged in! They will have a session with a non-null `user_id` when they are signed in

db/migrations/20240622213727_AddUserPasswordHash.up.sql
```diff
@@ -0,0 +1,3 @@
+-- Add migration script here
+ALTER TABLE Users
+ADD COLUMN password_hash TEXT NULL;
```
coreyja, line 3: We are storing the hash of passwords now that we are doing real auth!

db/migrations/20240623020146_AddSessionLoggedOutAt.up.sql
```diff
@@ -0,0 +1,3 @@
+-- Add migration script here
+ALTER TABLE Sessions
+ADD COLUMN logged_out_at TIMESTAMPTZ NULL;
```
coreyja, line 3: We also want to record when you logout of a session, so that we can treat it as not-active anymore

db/src/users.rs
```diff
@@ -6,6 +6,7 @@ use uuid::Uuid;
 pub struct User {
     pub user_id: Uuid,
     pub user_name: String,
+    pub password_hash: Option<String>,
```
coreyja: This is an `Option` since not all the existing users in your local DB will have a password_hash

But all Users we create from the web side are expected to have password_hashes!

Eventually we will likely do some sort of 'backfill' to get password hashes into all our DBs so that we can make this column non-null (and then remove this `Option`) here!

fly.toml
```diff
@@ -0,0 +1,22 @@
+# fly.toml app configuration file generated for knowledge-snowy-hill-4870 on 2024-06-22T12:26:58-04:00
+#
+# See https://fly.io/docs/reference/configuration/ for information about how to use this file.
+#
+
+app = 'knowledge-snowy-hill-4870'
```
coreyja, line 6: This file tells `fly` how to configure our app

We will also likely rebrand here, but didn't want to spend time on that while I was coding lol

web/src/cron.rs
```diff
@@ -0,0 +1,21 @@
...
+use crate::{jobs, AppState};
+
+fn hours(h: u64) -> Duration {
+    Duration::from_secs(h * 60 * 60)
+}
+
+fn cron_registry() -> CronRegistry<AppState> {
+    let mut registry = CronRegistry::new();
+
+    registry.register_job(jobs::sessions::Cleanup, hours(1));
```
coreyja, line 14: We only have a single job that we register on the Cron

Every hour we want to check the Sessions table for sessions that are more that 24 hours expired and delete them to keep the table nice and clean

web/src/err.rs
```diff
@@ -0,0 +1,34 @@
+use std::fmt::Display;
+
+use axum::response::IntoResponse;
+use tracing::error;
+
+#[derive(Debug)]
+pub struct Error {
```
coreyja, line 7: Don't worry _too_ much about this file, just some boilerplate to allow us to do some better errors on the Web side

web/src/jobs.rs
```diff
@@ -0,0 +1,3 @@
+pub mod sessions;
+
+cja::impl_job_registry!(crate::AppState, sessions::Cleanup);
```
coreyja, line 3: Just like in Cron we have to register our single Job here. As we add more jobs they will each need to be listed here so the system can 'find' them

web/src/jobs/sessions.rs
```diff
@@ -0,0 +1,28 @@
...
+            DELETE FROM sessions
+            WHERE expires_at < now() - interval '1 day'
+            "#,
+        )
+        .execute(db)
+        .await
+        .into_diagnostic()?;
+
+        Ok(())
+    }
```
coreyja, line 27: This is how we write jobs! We have a struct call `Cleanup` and then we implement the `Job<AppState>` trait

This required we implement on method called `run` where we want to logic of our job to live!

Here we simply delete old sessions!

web/src/jobs/sessions.rs
```diff
@@ -0,0 +1,28 @@
...
+
+        sqlx::query!(
+            r#"
+            DELETE FROM sessions
+            WHERE expires_at < now() - interval '1 day'
+            "#,
+        )
+        .execute(db)
+        .await
+        .into_diagnostic()?;
```
coreyja, line 24: Inside jobs you will need to use `into_diagnostic()` before you use `?`

This is cause this uses `miette` instead of `color_eyre`

web/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(4)
+        .enable_all()
+        .build()
+        .into_diagnostic()?
+        .block_on(async { _main().await })
+}
+
+async fn _main() -> miette::Result<()> {
```
coreyja, line 49: `_main` is just cause our error reporter service `Sentry` wants to be configured BEFORE we start up Tokio, just FYI

web/src/pages.rs
```diff
@@ -0,0 +1,30 @@
+use axum::response::{IntoResponse, Response};
```
coreyja, line 1: Just moved a few routes into here to organize!

web/src/pages.rs
```diff
@@ -0,0 +1,30 @@
...
+use db::users::User;
+
+use crate::templates::{Template, TemplatedPage};
+
+pub async fn home(t: Template, user: Option<User>) -> Response {
```
coreyja, line 6: The home route either renders the Landing page, or the Home page depending on if the User is logged in!

You can add `user: Option<User>` to your route handlers and Axum will automatically check if the User is logged in and give you a `Some(user)` if they are and a `None` otherwise!

web/src/pages.rs
```diff
@@ -0,0 +1,30 @@
...
+    match user {
+        Some(user) => user_dashboard(t, user).await.into_response(),
+        None => landing(t).await.into_response(),
+    }
+}
+
+pub async fn landing(t: Template) -> TemplatedPage {
```
coreyja, line 13: I added this `t: Template` thing instead of calling the `template` method like we were doing when we paird!

Its _basically_ the same thing. We want to write the 'main' part of the page here, but want the extra stuff like the HTML header and script tags to be added for us!

So now you add `t: Template` to your route handler, and then call `t.render` to wrap the html of the page!

web/src/pages.rs
```diff
@@ -0,0 +1,30 @@
...
+    t.render(maud::html! {
+        h1 { "Knowledge" }
+        h2 { "A cool app that needs a new name" }
+
+        p { "Welcome to Knowledge! Please log in." }
+        a href="/login" { "Login" }
+    })
+}
+
+pub async fn user_dashboard(t: Template, user: User) -> TemplatedPage {
```
coreyja, line 23: You can also add `user: User` without the wrapping option to your route handler!

And this time if the user is NOT logged in they will not be able to go to this route, and will instead be redirects to the `/login` page!

web/src/routes.rs
```diff
@@ -0,0 +1,17 @@
...
+pub fn routes(app_state: AppState) -> axum::Router {
+    axum::Router::new()
+        .route("/", get(home))
+        .route("/hello", get(landing))
+        .route("/dashboard", get(user_dashboard))
+        .route("/login", get(users::login::get).post(users::login::post))
+        .route("/signup", get(users::signup::get).post(users::signup::post))
+        .route("/logout", get(users::login::logout))
+        .with_state(app_state)
+}
```
coreyja, line 17: Here is where we define all the routes for our app!

web/src/routes.rs
```diff
@@ -0,0 +1,17 @@
...
+    pages::{home, landing, user_dashboard},
+    users, AppState,
+};
+
+pub fn routes(app_state: AppState) -> axum::Router {
+    axum::Router::new()
+        .route("/", get(home))
+        .route("/hello", get(landing))
+        .route("/dashboard", get(user_dashboard))
+        .route("/login", get(users::login::get).post(users::login::post))
```
coreyja, line 13: The `login` page accepts both a `GET` request and a `POST` request

The GET is what the browser sends when you go to the page, so this renders the login form for the user. When the user fills out the form and submits it that sends a POST request to the server with the values of the form!

So the `POST` here is where we take the user input and try to log them in!

web/src/sessions.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub session_id: Uuid,
+    pub user_id: Option<Uuid>,
+    pub expires_at: chrono::DateTime<chrono::Utc>,
+    pub logged_out_at: Option<chrono::DateTime<chrono::Utc>>,
+    pub updated_at: chrono::DateTime<chrono::Utc>,
+    pub created_at: chrono::DateTime<chrono::Utc>,
+}
+
+impl Session {
+    async fn create(state: &AppState, cookies: &Cookies) -> Result<Self, SessionError> {
```
coreyja, line 26: This method creates a Session in the DB, AND sets a browser cookie with its value so the next page load we remember who this user is!

web/src/sessions.rs
```diff
@@ -0,0 +1,115 @@
...
+pub struct SessionError;
+
+impl IntoResponse for SessionError {
+    fn into_response(self) -> Response<Body> {
+        http::StatusCode::INTERNAL_SERVER_ERROR.into_response()
+    }
+}
+
+#[async_trait::async_trait]
+impl FromRequestParts<AppState> for Session {
```
coreyja, line 67: This is how we let you add `session: Session` to your handlers and axum automatically knows what to do!

Well what it does is call this `from_request_parts` method to make a `Session` for you before calling your handler!

This code looks in the Users cookies for a `session_id` and uses that to pull the right session from the DB

web/src/templates.rs
```diff
@@ -0,0 +1,86 @@
...
+        let Query(query) =
+            Query::<HashMap<String, String>>::from_request_parts(parts, _state).await?;
+        let error = query.get("flash[error]").cloned();
+
+        Ok(Flash { error })
+    }
+}
+
+pub struct Template {
+    flash: Flash,
```
coreyja, line 32: Inside our template we have these little `flashes`. These are why we made the Template struct over the method we had when we paired

I wanted to be able to send a simple message from one page to the next to be displayed to the user as an error

So the tempalte now looks for a `?flash` query param that it can pull out and put in this `Flash` struct

When we render the template if there is an error flash we render it before the main page content! We use this for instance when you try to signup with an existing username

web/src/templates.rs
```diff
@@ -0,0 +1,86 @@
...
+            head {
+                meta charset="UTF-8";
+                meta name="viewport" content="width=device-width, initial-scale=1.0";
+                script src="https://cdn.tailwindcss.com" {}
+            }
+
+            body {
+                @if let Some(error) = &self.template.flash.error {
+                    p class="color-red-500" { (error) }
+                }
```
coreyja, line 75: This is where we actually render that flash if its present!

web/src/users/login.rs
```diff
@@ -0,0 +1,110 @@
...
+
+pub async fn get(t: Template) -> impl IntoResponse {
+    t.render(maud::html! {
+        form method="post" action="/login" {
+            input type="text" name="username" placeholder="Username";
+            input type="password" name="password" placeholder="Password";
+            input type="submit" value="Login";
+        }
+    })
+}
```
coreyja, line 25: Here is where we build the login page!

Its a very boring HTML form for now lol

web/src/users/login.rs
```diff
@@ -0,0 +1,110 @@
...
+
+#[derive(serde::Deserialize)]
+pub struct FromData {
+    username: String,
+    password: String,
+}
+
+#[axum::debug_handler(state = AppState)]
+pub async fn post(
```
coreyja, line 34: And this is what gets called when you submit that form!

We get the `FormData` (typoed as FromData here lol) by telling Axum the user filled out a form and we want the data to fit into our `FromData` struct!

We also need the `Session` and the `AppState` so we can record the user as logged in!


We first look up the user they asked for by username, then we verify if the password is correct. If it is NOT we redirect back to login with a flash message.
If the password is valid we save the `user_id` to the `Session` which logs the user in!

web/src/users/login.rs
```diff
@@ -0,0 +1,110 @@
...
+    .await?;
+
+    Ok(Redirect::to("/dashboard"))
+}
+
+pub async fn logout(
+    session: Session,
+    State(state): State<AppState>,
+    cookies: Cookies,
+) -> WebResult<Redirect> {
```
coreyja, line 93: When you logout we want to remove the cookie from your browser, and also mark when you logged out in the DB

web/src/users/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn from(_: SessionError) -> Self {
+        ExtractUserError::NoSession
+    }
+}
+
+#[async_trait::async_trait]
+impl FromRequestParts<AppState> for User {
+    type Rejection = ExtractUserError;
+
+    async fn from_request_parts(
```
coreyja, line 51: This is what allows us to ask for a `user: User` in our handlers!

We take the session and look for a `user_id` to load

web/src/users/signup.rs
```diff
@@ -0,0 +1,70 @@
...
+
+pub async fn get(t: Template) -> impl IntoResponse {
+    t.render(maud::html! {
+      form method="post" action="/signup" {
+        input type="text" name="username" placeholder="Username";
+        input type="password" name="password" placeholder="Password";
+        input type="submit" value="Signup";
+      }
+    })
+}
```
coreyja, line 21: Signup form just like login!

web/src/users/signup.rs
```diff
@@ -0,0 +1,70 @@
...
+pub struct FormData {
+    username: String,
+    password: String,
+}
+
+pub async fn post(
+    session: Session,
+    State(state): State<AppState>,
+    form_data: Form<FormData>,
+) -> WebResult<Response> {
```
coreyja, line 33: And here we create a new user instead of logging one in!

We DO also log you in after you signup

Seif-Mamdouh (approved):
Wow... definitely learned a lot

web/src/users/login.rs
```diff
@@ -53,7 +53,11 @@ pub async fn post(
     };
 
     let Some(password_hash) = potential_user.password_hash.as_ref() else {
-        return Ok(Redirect::to("/login"));
+        tracing::error!(
+            user_id = potential_user.user_id.to_string(),
+            "User has no password set"
+        );
+        return Ok(Redirect::to("/login?flash[error]=No password set for this user"));
```
Seif-Mamdouh: This functions renders a error if the username is in the DB W/O a password. Now that I think about it, do we need this condition anymore since `username` && `password` are not `nil`?

db/migrations/20240624201152_AddPasswordHashNotNull.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users 
+DROP COLUMN password_hash;
```
Seif-Mamdouh: updated the migration to make the `password_hash` not nil

db/src/users.rs
```diff
@@ -29,7 +29,8 @@ pub async fn get_user(pool: &PgPool) -> Result<Vec<User>> {
 
 pub async fn get_username_by_id(pool: &PgPool, user_id: Uuid) -> Result<String> {
     let record = sqlx::query!("SELECT user_name FROM Users WHERE user_id = $1", user_id)
-        .fetch_optional(pool)
+        .fetch_one(pool)
         .await?;
-    Ok(record.map(|r| r.user_name).unwrap_or_default())
+
+    Ok(record.user_name)
```
Seif-Mamdouh: The update was to switch handling from `Option` type to `String` type returned by the `fetch_one` method.

web/src/jobs/sessions.rs
Seif-Mamdouh: `Job <AppState>` is imported from the `cja` crate correct?

Seif-Mamdouh: If so, where can I find the code for `App State` trait?

web/src/pages.rs
Seif-Mamdouh: funny looking back at this now 😂. Glad we made the change

web/src/sessions.rs
Seif-Mamdouh: Ohhh that's sick! Does that mean the user doesn't have to re-login after leaving the page and coming back?

web/src/sessions.rs
Seif-Mamdouh: Is this polymorphism? Also what's the reason for creating a new `instance` of Session instead of making another `async `function inside the `Session Trait`? Just curious 😆

web/src/templates.rs
Seif-Mamdouh: Does that have to do anything with our `flash` crate?

db/src/users.rs
coreyja: Ah I think this change for the `Option` probably wanted to stay so I reverted it!

This option wasnt really about the null-ness of the Password hash, its the difference between the `fetch_once` and `fetch_optional`, and the usages of this method in the CLI already handled the option, so I left it as is!

db/migrations/20240624201152_AddPasswordHashNotNull.down.sql
coreyja: I edited this migration so that instead of dropping the columm here we set it to nullable!

Because the idea of the `down` migration is to reverse the `up`, and since the up didn't create the column we dont want to drop it here!

web/src/sessions.rs
coreyja: No this isn't really polymorphism, I think Axum calls these `extractors`. Basically we can 'extract' some data from the requests (and in our case the DB helps too!)

> Also what's the reason for creating a new instance of Session instead of making another async function inside the Session Trait? 

First just some semantics! `Session` here is a struct, and the `FromRequestParts` is the trait we are implementing for our `Session` struct!

So we want an instance of a Session struct so that it can have all its data! And we need an instance of the struct to call methods on it that take a `self` parameter, or just to access the data they hold!

And in our request handlers, we want to know what specific Session this web request is associated with. We care about the _specific_ session_id not just the general concept of a Session, so we want an instance!

Does that kinda make sense?

web/src/templates.rs
coreyja: We don't have a crate for `flash` but it does related to the `Flash` struct I added!

The `Flash` struct is used to old the query params I mention here once they are de-serialized!

PR #17, opened 2024-06-25
Ticket 293 ability to add an article from the web UI

Seif-Mamdouh:
In this PR, I added an `HTML form` that allows users to submit a URL.

This feature meets the requirements outlined in Linear Ticket ALEX-293: https://linear.app/lavender-iguana/issue/ALEX-293/ability-to-add-an-article-from-the-web-ui

Which specifies the ability to add a row in the DB from the web UI.

Here's a quick demo:

https://github.com/coreyja/knowledge/assets/55955558/e2088083-3607-4c73-b3de-f90eeee2a642

### DIFF
--- a/web/src/pages.rs
+++ b/web/src/pages.rs
@@ -28,9 +28,9 @@ pub async fn user_dashboard(t: Template, user: User) -> TemplatedPage {
         a href="/logout" { "Logout" }
 
         h3 {  "Insert Article" }
-        form method="post" action="/insert_article" {
-            label for="url" { "URL:" }
-            textarea id="url" name="url" placeholder="Insert URL" {}
+        form method="post" action="/articles" {
+            label for="url" { "URL :" }
+            input id="url" name="url" placeholder="Insert URL" {}
             br;
             input type="submit" value="Submit";
         }
--- a/web/src/pages.rs
+++ b/web/src/pages.rs
@@ -29,8 +29,8 @@ pub async fn user_dashboard(t: Template, user: User) -> TemplatedPage {
 
         h3 {  "Insert Article" }
         form method="post" action="/insert_article" {
-            label for="content" { "Content:" }
-            textarea id="content" name="content" placeholder="Article Content" {}
+            label for="url" { "URL:" }
+            textarea id="url" name="url" placeholder="Insert URL" {}
             br;
             input type="submit" value="Submit";
         }
--- a/web/src/pages.rs
+++ b/web/src/pages.rs
@@ -28,7 +28,7 @@ pub async fn user_dashboard(t: Template, user: User) -> TemplatedPage {
         a href="/logout" { "Logout" }
 
         h3 {  "Insert Article" }
-        form method="post" action="/insert_articles" {
+        form method="post" action="/articles" {
             label for="url" { "URL :" }
             input id="url" name="url" placeholder="Insert URL" {}
             br;
--- a/web/src/pages.rs
+++ b/web/src/pages.rs
@@ -26,5 +26,13 @@ pub async fn user_dashboard(t: Template, user: User) -> TemplatedPage {
         p { "Welcome, " (user.user_name) }
 
         a href="/logout" { "Logout" }
+
+        h3 {  "Insert Article" }
+        form method="post" action="/insert_article" {
+            label for="content" { "Content:" }
+            textarea id="content" name="content" placeholder="Article Content" {}
+            br;
+            input type="submit" value="Submit";
+        }
     })
 }
--- a/web/src/pages.rs
+++ b/web/src/pages.rs
@@ -28,7 +28,7 @@ pub async fn user_dashboard(t: Template, user: User) -> TemplatedPage {
         a href="/logout" { "Logout" }
 
         h3 {  "Insert Article" }
-        form method="post" action="/articles" {
+        form method="post" action="/insert_articles" {
             label for="url" { "URL :" }
             input id="url" name="url" placeholder="Insert URL" {}
             br;
--- a/web/src/routes.rs
+++ b/web/src/routes.rs
@@ -1,8 +1,9 @@
 use axum::routing::get;
+use axum::routing::post;
 
 use crate::{
     pages::{home, landing, user_dashboard},
-    users, AppState,
+    users, AppState
 };
 
 pub fn routes(app_state: AppState) -> axum::Router {
@@ -13,5 +14,6 @@ pub fn routes(app_state: AppState) -> axum::Router {
         .route("/login", get(users::login::get).post(users::login::post))
         .route("/signup", get(users::signup::get).post(users::signup::post))
         .route("/logout", get(users::login::logout))
+        .route("/insert_article", post(users::add_url::insert_article_handler))
         .with_state(app_state)
 }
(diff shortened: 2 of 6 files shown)


web/src/routes.rs
```diff
@@ -13,5 +14,9 @@ pub fn routes(app_state: AppState) -> axum::Router {
         .route("/login", get(users::login::get).post(users::login::post))
         .route("/signup", get(users::signup::get).post(users::signup::post))
         .route("/logout", get(users::login::logout))
+        .route(
+            "/insert_article",
+            post(users::add_url::insert_article_handler),
+        )
         .with_state(app_state)
```
Seif-Mamdouh, line 18: we added route called "/insert_article" and it sends the outcome of the insert_article_handler to the POST request to the "/insert_article" endpoint.

web/src/users/add_url.rs
```diff
@@ -0,0 +1,32 @@
...
+pub async fn insert_article_handler(
+    State(state): State<AppState>,
+    user: User,
+    Form(form): Form<ArticleForm>,
+) -> impl IntoResponse {
+    let url = form.content;
+    let user_id = user.user_id;
+
+    println!("inserting");
+    match add_url(&state.db, &url, user_id, &true).await {
```
Seif-Mamdouh, line 26: Attempt to add URL to the DB

web/src/users/add_url.rs
```diff
@@ -0,0 +1,32 @@
...
+use serde::Deserialize;
+
+use crate::AppState;
+
+#[derive(Deserialize)]
+pub struct ArticleForm {
+    content: String,
+}
+
+#[axum::debug_handler(state = AppState)]
```
Seif-Mamdouh, line 16: This function is marked with `#[axum::debug_handler]` to enable debugging and state management.

web/src/users/add_url.rs
```diff
@@ -0,0 +1,32 @@
...
+pub async fn insert_article_handler(
+    State(state): State<AppState>,
+    user: User,
+    Form(form): Form<ArticleForm>,
+) -> impl IntoResponse {
+    let url = form.content;
+    let user_id = user.user_id;
+
+    println!("inserting");
```
coreyja: Instead of using `println` lets start using the `info!` and `debug!` and `error!` traits from `tracing`!
https://github.com/tokio-rs/tracing#in-applications

`tracing` is kinda like a logging and more platform for Rust!
For now basically just replace `println!` with `info!` and `eprintln!` with `error!`
but we'll def do more tracing as we get going!

web/src/users/add_url.rs
```diff
@@ -0,0 +1,32 @@
...
+    response::{IntoResponse, Redirect},
+};
+use db::{urls::add_url, users::User};
+use serde::Deserialize;
+
+use crate::AppState;
+
+#[derive(Deserialize)]
+pub struct ArticleForm {
+    content: String,
```
coreyja: Can we rename this field to be `url` or something? And also change the form label and stuff?

At first I wasn't sure if the form wanted me to like copy paste the article body and I was a bit confused 😆

web/src/users/add_url.rs
```diff
@@ -0,0 +1,32 @@
...
+    State(state): State<AppState>,
+    user: User,
+    Form(form): Form<ArticleForm>,
+) -> impl IntoResponse {
+    let url = form.content;
+    let user_id = user.user_id;
+
+    println!("inserting");
+    match add_url(&state.db, &url, user_id, &true).await {
+        Ok(_) => Redirect::to("/dashboard"),
```
coreyja: Nice! Eventually I think this might redirect to a page about the specific article we added but we can do that next 💯

web/src/routes.rs
```diff
@@ -13,5 +14,9 @@ pub fn routes(app_state: AppState) -> axum::Router {
         .route("/login", get(users::login::get).post(users::login::post))
         .route("/signup", get(users::signup::get).post(users::signup::post))
         .route("/logout", get(users::login::logout))
+        .route(
+            "/insert_article",
```
coreyja: I think `insert` might be able to be removed! Since we are using a `POST` request its kinda implied that this is a create operation, so we don't really need to have it in the URL too 🤷 

And that way we can use the `/articles` `GET` later for a page that returns all the articles you saved or something!

Oh and can we do plural routes? So `articles` instead of `article`? Not a big deal just a convention lol

web/src/users/add_url.rs
coreyja: And actually since this is a redirect to the page you are already on, want to add a flash message here too?

Might need to extend the Flash stuff to have a `success` type instead of the error. Up for giving it a shot?

web/src/pages.rs
```diff
@@ -26,5 +26,13 @@ pub async fn user_dashboard(t: Template, user: User) -> TemplatedPage {
         p { "Welcome, " (user.user_name) }
 
         a href="/logout" { "Logout" }
+
+        h3 {  "Insert Article" }
+        form method="post" action="/insert_article" {
+            label for="content" { "Content:" }
+            textarea id="content" name="content" placeholder="Article Content" {}
```
coreyja: SuperNit: A `textarea` is 'bigger' and usually for like long form content

For just this url I might make this a standard `input`!

coreyja (changes requested):
Looks great! Left some comments

web/src/users/add_url.rs
Seif-Mamdouh: That's very helpful for debugging!

web/src/users/add_url.rs
Seif-Mamdouh: Yea absolutely!

web/src/routes.rs
Seif-Mamdouh: Roger that 🫡

web/src/templates.rs
```diff
@@ -10,6 +10,7 @@ use crate::{err, AppState};
 
 pub struct Flash {
     error: Option<String>,
+    success: Option<String>,
 }
```
Seif-Mamdouh, line 14: added a `success` data type in this flash struct

web/src/routes.rs
```diff
@@ -13,5 +14,9 @@ pub fn routes(app_state: AppState) -> axum::Router {
         .route("/login", get(users::login::get).post(users::login::post))
         .route("/signup", get(users::signup::get).post(users::signup::post))
         .route("/logout", get(users::login::logout))
+        .route(
+            "/insert_articles",
```
coreyja:
```suggestion
            "/articles",
```

web/src/pages.rs
```diff
@@ -26,5 +26,13 @@ pub async fn user_dashboard(t: Template, user: User) -> TemplatedPage {
         p { "Welcome, " (user.user_name) }
 
         a href="/logout" { "Logout" }
+
+        h3 {  "Insert Article" }
+        form method="post" action="/insert_articles" {
```
coreyja: Ah so we made this plural, but lets also drop the `insert` prefix!

```suggestion
        form method="post" action="/articles" {
```

coreyja (approved):
Awesome!

I think we got half of the route renaming, and I'm gonna make two quick suggestion edits here and then merge this in!

web/src/routes.rs
```diff
@@ -13,5 +14,9 @@ pub fn routes(app_state: AppState) -> axum::Router {
         .route("/login", get(users::login::get).post(users::login::post))
         .route("/signup", get(users::signup::get).post(users::signup::post))
         .route("/logout", get(users::login::logout))
+        .route(
+            "/articles",
+            post(users::add_url::insert_article_handler),
+        )
```
coreyja:
```suggestion
        .route("/articles", post(users::add_url::insert_article_handler))
```

PR #18, opened 2024-06-28
Linear ticket 295 process article using a job

Seif-Mamdouh:
### Summary
In this PR we added changes to add a job for processing articles that User has inserted in the Web UI and displaying the summary in the web UI.  🎉 

### Key Changes
- added a new job for processing articles in `web/src/jobs/process_article.rs`
- added new route to display the summary of an article on the web page
- updated the `PageSnapShot Table` to add a `summary` column

### Couple of thoughts while building this

1. In the `PageSnapShot` table, we only added a `summary` column. However, I was wondering how we can attach the `category` to the related `summary`. 
Currently, a singular `category` is generated through a `markdown_id` in the `Category` table, and the way we generate a `summary` is through `page_snapshot_id`. To solve this I was thinking we could add a `category` column to the `PageSnapShot` table or move the `summary` column to the Markdown table since both `summary` and `category` will share the same `markdown_id`.
Happy to chat about it more! 😁 

2. Whenever the user adds a URL, if it's successful it renders a `No page snapshot found`. I find it very kind of misleading, I was wondering maybe we could add a another message such as maybe `processing article ....` or maybe even add a loading state! 

3. To see the summary for a article, I have to refresh a couple of times to see the summary on the Web UI. Would love to discuss on how can I fix this without having to refresh the page. 

### Quick Demo

https://github.com/coreyja/knowledge/assets/55955558/caff589d-a4d4-4e38-9fdb-f9669da667f1


db/src/urls.rs
```diff
@@ -139,7 +149,26 @@ async fn store_markdown(
     Ok(markdown_result)
 }
 
-async fn store_raw_html_in_page_snapshot(
+pub async fn generate_and_store_summary(
```
Seif-Mamdouh: Added a `generate_and_store_summary` function to insert summary in the `PageSnapShot Table`.

db/src/urls.rs
```diff
@@ -160,14 +189,15 @@ async fn store_raw_html_in_page_snapshot(
     .fetch_one(pool)
     .await?;
 
-    let markdown_result = store_markdown(pool, result.page_snapshot_id, &cleaned_html).await?;
-    println!("Markdown result: {markdown_result:?}");
+    store_markdown(pool, result.page_snapshot_id, &cleaned_html).await?;
+    let markdown_content = html2md::parse_html(&cleaned_html);
+    generate_and_store_summary(pool, result.page_snapshot_id, &markdown_content).await?;
```
Seif-Mamdouh: calling the `generate_and_store_summary` to generate a article summary related to `page_snapshot_id`

web/src/jobs/process_article.rs
```diff
@@ -0,0 +1,33 @@
...
+            .into_diagnostic()?;
+
+        // store the html content in the page snapshot
+        let page_snapshot = store_html_content_in_page_snapshot(db, page.clone()).await;
+
+        println!("{:?}", page_snapshot);
+
+        Ok(())
+    }
+}
```
Seif-Mamdouh, line 56: The `ProcessArticle` job retrieves a page from the database using its and stores the `HTML content` in a `PageSnapShot`

web/src/jobs/process_article.rs
```diff
@@ -0,0 +1,33 @@
...
+
+        let page = sqlx::query_as!(Page, "SELECT * FROM pages WHERE page_id = $1", self.page_id)
+            .fetch_one(db)
+            .await
+            .into_diagnostic()?;
+
+        // store the html content in the page snapshot
+        let page_snapshot = store_html_content_in_page_snapshot(db, page.clone()).await;
+
+        println!("{:?}", page_snapshot);
```
Seif-Mamdouh: I'm printing this out because clippy was yelling at me 😂. 
Screenshot 2024-06-28 at 10 38 58 AM
If I call a function `store_html_content_in_page_snapshot` clippy thinks I'm not using it anywhere. Would love to get some insights on how to fix this!

web/src/routes.rs
```diff
@@ -18,5 +19,6 @@ pub fn routes(app_state: AppState) -> axum::Router {
             "/insert_articles",
             post(users::add_url::insert_article_handler),
         )
+        .route("/article/:article_id", get(article_detail))
```
Seif-Mamdouh: new route to display the summary of an article on the web page

web/src/pages.rs
```diff
@@ -36,3 +47,31 @@ pub async fn user_dashboard(t: Template, user: User) -> TemplatedPage {
...
+
+    Ok(t.render(maud::html! {
+        @if let Some(page_snapshot) = page_snapshot {
+            p { (page_snapshot.summary) }
+        } @else {
+            p { "No page snapshot found" }
+        }
+    })
+    .into_response())
+}
```
Seif-Mamdouh, line 92: Function `article_detail` handles the inserted `URL` by the User, Enqueues the `Process Article` and redirect the response indicating `success` or `error`

db/Cargo.toml
```diff
@@ -18,6 +18,7 @@ sqlx = { workspace = true, features = ["postgres", "runtime-tokio-rustls", "chro
 tracing = { workspace = true }
 url.workspace = true
 uuid = { workspace = true }
+cja = { version = "0.0.0", git = "https://github.com/coreyja/coreyja.com", branch = "main" }
```
coreyja: Does `db` need this dep?

`cja` is for more of the web side of things, so ideally the `db` crate doesn't need to depend on it, and just the `web` crate can

If this _is_ needed do you think we can move whatever logic that touches `cja` 'up' to the web crate?

db/migrations/20240628024728_AddSummaryToPageSnapShot.up.sql
```diff
@@ -0,0 +1,2 @@
+-- Add migration script here
+ALTER TABLE PageSnapshot ADD COLUMN summary TEXT DEFAULT '' NOT NULL;
```
coreyja: Ahh ya I like that you said in the PR Description!

Let's add this to `Markdown` where the category and everything is instead of the PageSnapshot I like that!

db/src/urls.rs
```diff
@@ -24,13 +24,23 @@ pub enum AddUrlOutcome {
     Existing(Page),
 }
 
+impl AddUrlOutcome {
+    #[must_use]
```
coreyja: I _think_ we can probably remove this `#[must_use]`

Things like `Result` are marked like that because Rust doesn't want people to ignore the `Err` variant and just be throwing away errors

But this `Page` reference isn't 'scary' to forget to use, so I don't we need this!

Was there a specific reason for adding it?

db/src/urls.rs
```diff
@@ -114,14 +124,14 @@ async fn store_category(
...
+pub async fn store_markdown(
     pool: &PgPool,
     page_snapshot_id: Uuid,
     cleaned_html: &str,
 ) -> color_eyre::Result<Markdown> {
     let markdown_content = html2md::parse_html(cleaned_html);
-    let summary = generate_summary(&markdown_content).await?;
-    println!("Summary: {summary}");
+    // let summary = generate_summary(&markdown_content).await?;
+    // println!("Summary: {summary}");
```
coreyja: Want to just delete these if we don't need them anymore?

db/src/urls.rs
```diff
@@ -139,7 +149,26 @@ async fn store_markdown(
...
+        summary,
+        page_snapshot_id
+    )
+    .execute(pool)
+    .await?;
+
+    Ok(())
+}
+
+pub async fn store_html_content_in_page_snapshot(
```
coreyja: This fn does a good bit more than this name suggests 🤔 

Wonder if we want a function that _just  does the HTML in PageSnapshot_ or if we want to rename this message to something more "high level" like `persist_article` or something

Also relatedly we should probably decide on some terminology and be consistent with it through the App (and UI!)

Like do we call things articles? Or Pages? Should the PageTable be an 'Article' Table?

Not a _huge_ deal for now, we can easily rename stuff. But good to think about!

db/src/urls.rs
coreyja: So here we call `store_markdown` and then `generate_and_store_summary` but I _believe_ both of these actually take the HTML and make the markdown.

And it also looks like `store_markdown` might return a Struct with the markdown!

Could we pass something different into `generate_and_store_summary` so that it already has the markdown and doesn't need to re-parse it?

web/src/jobs/process_article.rs
```diff
@@ -0,0 +1,33 @@
...
+use db::urls::{store_html_content_in_page_snapshot, Page};
+
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+pub struct ProcessArticle {
+    pub page_id: Uuid,
+}
+
+#[async_trait::async_trait]
+impl Job<AppState> for ProcessArticle {
+    const NAME: &'static str = "process_article::processArticle";
```
coreyja: SuperNit: Lets make this match the struct casing (even though I think I typed this out with you live lol)

```suggestion
    const NAME: &'static str = "process_article::ProcessArticle";
```

web/src/jobs/process_article.rs
coreyja: Ahh! So this is actually coming from the `#[must_use]` attribute that Rust puts on `Result` 😆 

So `page_snapshot` here is a `Result` since we don't have a `.into_diagnostic()?` at the end of the line to forward that up the the caller of `run`

But I think if we add the `?` then we can remove the `println!` cause Rust won't be upset if we don't use that value, just the `Result` since its special

web/src/jobs/process_article.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    async fn run(&self, app_state: AppState) -> miette::Result<()> {
+        let db = app_state.db();
+
+        let page = sqlx::query_as!(Page, "SELECT * FROM pages WHERE page_id = $1", self.page_id)
+            .fetch_one(db)
+            .await
+            .into_diagnostic()?;
+
+        // store the html content in the page snapshot
```
coreyja: Nit: I'd remove this comment, its basically repeating the method name so I don't think it provides a lot of long term value to the code base!

web/src/jobs/process_article.rs
coreyja: SuperNit: I might but these comments on the line where we declare the function

And thats mostly cause of how we read code from top to bottom most of the time. By the time I make it to the end of the function, I've already kinda figured out what its doing lol

But this as an 'intro' to the function before I dive into the code would be super helpful!

web/src/pages.rs
```diff
@@ -26,5 +37,41 @@ pub async fn user_dashboard(t: Template, user: User) -> TemplatedPage {
         p { "Welcome, " (user.user_name) }
 
         a href="/logout" { "Logout" }
+
+        h3 {  "Insert Article" }
+        form method="post" action="/insert_articles" {
```
coreyja: This is actually a merge/rebase issue kinda

I changed this url to `/articles` before I merged your last PR so it will likely need to be updated here too!

Change in last PR: [`fff0454` (#17)](https://github.com/coreyja/knowledge/pull/17/commits/fff04544b6e546dc7c0259d662a8f57e194b9620)

web/src/pages.rs
coreyja: Not a big deal but that's not quite right is it? `article_detail` looks to be the handler for when we are viewing an article in the UI, right?

This handler looks great! Just making sure we are on the same page lol

web/src/pages.rs
```diff
@@ -26,5 +37,41 @@ pub async fn user_dashboard(t: Template, user: User) -> TemplatedPage {
...
+        article_id.page_id
+    )
+    .fetch_optional(&state.db)
+    .await?;
+
+    Ok(t.render(maud::html! {
+        @if let Some(page_snapshot) = page_snapshot {
+            p { (page_snapshot.summary) }
+        } @else {
+            p { "No page snapshot found" }
```
coreyja: From your PR description comment, ya we can/should totally change this to a `Snapshot generating, please try again...` or something message!

And yes we will get to making this something more like a loading spinner, but we'll do that as another slice/card!

web/src/routes.rs
coreyja: I used `/articles/:article_id` instead to match the new route of the post above?

web/src/templates.rs
```diff
@@ -23,8 +24,9 @@ impl FromRequestParts<AppState> for Flash {
         let Query(query) =
             Query::<HashMap<String, String>>::from_request_parts(parts, _state).await?;
         let error = query.get("flash[error]").cloned();
+        let success = query.get("flash[success]").cloned();
```
coreyja: 💯  Perfect!

web/src/templates.rs
```diff
@@ -73,6 +75,9 @@ impl TemplatedPage {
                 @if let Some(error) = &self.template.flash.error {
                     p class="color-red-500" { (error) }
                 }
+                @if let Some(success) = &self.template.flash.success {
+                    p class="color-green-500" { (success) }
```
coreyja: We'll do some brand coloring soon!

But this is perfect lol

web/src/users/add_url.rs
```diff
@@ -0,0 +1,41 @@
...
+) -> impl IntoResponse {
+    let url = form.url;
+    let user_id = user.user_id;
+
+    match add_url(&state.db, &url, user_id, &true).await {
+        Ok(page) => {
+            let page_id = page.page().page_id;
+            let process_article = ProcessArticle { page_id };
+            process_article
+                .enqueue(state.clone(), "context".to_string())
```
coreyja: Ahh ya I never really explained what `context` should be!

So `context` is basically just a way for us a devs to be able to trace these jobs better.
In a big app we may be proccessing articles from all over the place. Like one might be this HTML form, and another might be from a Slack/Discord integration (that's likely coming up soonish 😉 )
And it's nice to be able to know 'where' jobs are coming from when are looking into the system!

So for this context... Lets use... `insert_article_handler` I think just to match the function name.
Another alternative is something more like `POST /articles` to match the route... 🤷 
No strong feels, but lets def make this more about the context of what we are doing/where we are in code!

web/src/users/add_url.rs
```diff
@@ -0,0 +1,41 @@
...
+                .await
+                .unwrap();
+            Redirect::to(&format!("/article/{page_id}?flash[success]=Article added"))
```
coreyja: Love this!

Just chatting, I think we might want some 'easier' way to make url paths like this in the future. Cause I kinda hate having to do it as string interpolation, but for now this is perfect!
This is mostly a note to myself to think about thinks to build into `cja` more!

web/src/users/add_url.rs
```diff
@@ -0,0 +1,41 @@
...
+        }
+        Err(e) => {
+            eprintln!("Error adding URL: {e:?}");
+            Redirect::to("/dashboard?flash[error]=Could not add article")
```
coreyja: Ok so I think we want to change this here, and unfortunately we are going to lose this nice redirect and flash... But it will lean into the existing Error Reporting better, and we will add back in nice redirects and flash messages later I think!

So I think what we want to do here is make this function return a `WebResult<Redirect>` (or `WebResult<impl IntoResponse>` if we don't want to be specific about the redirect, will compile the same just means if we change it from a redirect we won't have to change the signature 🤷 )

And then we should be able to use `?` on the `Result`s in this function to return a generic `500` error if they fail, which also handles reporting the error with the `error!` tracing macro: https://github.com/coreyja/knowledge/blob/main/web/src/err.rs#L30

And then you don't have to do the `match` here for the error, you can just use `?` to forward that up and have that `Error` renderer I linked run!

And can we do the same for the `unwrap()` after the enqueue? I _think_ that will work for both a `miette` error and `color_eyre` error (and its on my list to fix this likely my just moving to `color_eyre` in cja soon!)

web/src/routes.rs
```diff
@@ -13,5 +15,10 @@ pub fn routes(app_state: AppState) -> axum::Router {
         .route("/login", get(users::login::get).post(users::login::post))
         .route("/signup", get(users::signup::get).post(users::signup::post))
         .route("/logout", get(users::login::logout))
+        .route(
+            "/insert_articles",
```
coreyja: Reminder that this got changed to `/articles` and needs to be updated here too (or may when you rebase not positive)

[`fff0454` (#17)](https://github.com/coreyja/knowledge/pull/17/commits/fff04544b6e546dc7c0259d662a8f57e194b9620)

coreyja (changes requested):
Looking really good!

I think most of the my comments here are really about organization/code cleanliness, the functionality here is great!

coreyja:
Oh and a thought on PR description, but its really a nitpick but thought I'd share!

Similiar to leaving the comments at the top of a function, I might put your Thoughts ABOVE the demo. Really cause I think thats more important than the gif (which I really do appreciate tho!), and a little bit cause the demo takes me out of 'reading' mode lol

Cause I really like those thoughts! Love that you are thinking about those, and I think I responded to them inline in various comments above!

web/src/jobs/process_article.rs
Seif-Mamdouh: Roger that!

web/src/jobs/process_article.rs
Seif-Mamdouh: added `.unwrap()`  to `store_html_content_in_page_snapshot(db, page.clone()).await`

web/src/routes.rs
Seif-Mamdouh: I reason why I have `insert_articles` is because clippy is yelling at me for amount of characters is not satisfying for it 😂

db/Cargo.toml
Seif-Mamdouh: I added this cause I thought I would call the job into the `db` folder! Turned out I was wrong lol. I will remove it

db/src/urls.rs
Seif-Mamdouh: I think we added it while we pairing because `clippy` was yelling at us

db/src/urls.rs
Seif-Mamdouh: Yes we don't need it!

db/src/urls.rs
Seif-Mamdouh: You are correct! We should decide on the `naming terminology`. I say we go with article. I think it makes sense for me

web/src/users/add_url.rs
Seif-Mamdouh: Awesome! I will change it to `insert_article_handler` for now to know that we are processing `HTML` articles

web/src/users/add_url.rs
Seif-Mamdouh: Alright awesome! Any idea on how we can make this easier?

web/src/users/add_url.rs
Seif-Mamdouh: Sick! I changed it

What's the reason behind this change? Is the only reason because we want the a generic `500 error ` response?

Just curious 🧐

cli/src/main.rs
```diff
@@ -62,11 +67,15 @@ async fn main() -> color_eyre::Result<()> {
             allow_existing,
         } => {
```
Seif-Mamdouh, line 68: variables `user_id,`` raw_html`, `url_parsed`, and `cleaned_html` are intermediate steps required to process the URL, clean its HTML content, and store it in the database

cli/src/main.rs
```diff
@@ -62,11 +67,15 @@ async fn main() -> color_eyre::Result<()> {
...
             let user_id = get_user_id_from_session()?;
```
Seif-Mamdouh, line 67: variables `user_id,`` raw_html`, `url_parsed`, and `cleaned_html` are intermediate steps required to process the URL, clean its HTML content, and store it in the database

db/src/urls.rs
```diff
@@ -166,7 +166,11 @@ pub async fn generate_and_store_summary(
     Ok(())
 }
 
-pub async fn persist_article(pool: &PgPool, page: Page, markdown: Markdown) -> color_eyre::Result<PageSnapShot> {
+pub async fn persist_article(
```
Seif-Mamdouh: This ensures that the HTML is properly formatted and the URL is valid before being saved as a PageSnapShot.

Seif-Mamdouh:
One more thought while I was building this.

Every time we add a `URL` to the `DB`. We call `Markdown_id` from the `store markdown` function. This means if the User inserts the same URL, it will render a 500 Error since `Markdown_id` is a unique `PK` and cannot be added twice to the `DB`.

My first instinct is to make a `upsert function` however I wanted to chat about first 😃

db/src/urls.rs
```diff
@@ -136,12 +145,31 @@ async fn store_markdown(
     let category_result = store_category(pool, markdown_result.markdown_id, &category).await?;
     println!("Category result: {category_result:?}");
 
-    Ok(markdown_result)
+    Ok((markdown_result, markdown_content))
```
coreyja: Doesn't `markdown_result` already have the content at `markdown_result.markdown_content`? Do we need to pass is down again?

db/src/urls.rs
```diff
@@ -160,14 +188,18 @@ async fn store_raw_html_in_page_snapshot(
...
 
-pub async fn process_page_snapshot(pool: &PgPool, page: Page) -> color_eyre::Result<()> {
-    let outcome = store_raw_html_in_page_snapshot(pool, page).await?;
-    println!("Outcome: {outcome:?}");
+pub async fn process_page_snapshot(
+    pool: &PgPool,
+    page: Page,
+    markdown: Markdown,
+) -> color_eyre::Result<()> {
+    let _outcome = persist_article(pool, page, markdown).await?;
```
coreyja: I wonder if it would be better to just call `persist_article` inline?

Just cause that method is a bit more descriptive than `process_page_snapshot` I feel like!

coreyja: Or maybe what would be better is to take these bits of `persist_article` and move them into this method!

```
 store_markdown(pool, result.page_snapshot_id, &cleaned_html).await?;
    generate_and_store_summary(pool, result.page_snapshot_id, &markdown_content).await?;
```

That way the `persist_artcile` method doesn't _also_ have to create the summary and persist it!

Overall though I might still remove `process_page_snapshot` just cause the things it calls are nicely named, and I don't think this function is really helping us much 🤷

db/src/urls.rs
```diff
@@ -136,12 +145,31 @@ async fn store_markdown(
...
+    .execute(pool)
+    .await?;
+
+    Ok(())
+}
+
+pub async fn persist_article(
     pool: &PgPool,
     page: Page,
+    _markdown: Markdown,
```
coreyja: Do we need to be passing down this `Markdown`? Aren't we making the markdown inside this method?

web/src/jobs/process_article.rs
```diff
@@ -0,0 +1,40 @@
...
+            .into_diagnostic()?;
+
+        let markdown = sqlx::query_as!(
+            Markdown,
+            "SELECT * FROM markdown WHERE markdown_id = $1",
+            self.markdown_id
+        )
+        .fetch_one(db)
+        .await
+        .into_diagnostic()?;
```
coreyja: We don't expect to _have_ a markdown at this point do we?

This is before one would be generated right?

web/src/routes.rs
coreyja: Can we do `/artciles/:article_id` just to match the `/articles` plural route above this?

web/src/users/add_url.rs
```diff
@@ -18,16 +26,39 @@ pub async fn insert_article_handler(
...
+    let cleaned_html = clean_raw_html(&raw_html, &url_parsed)?;
+
+    let (markdown, _) = store_markdown(&state.db, page_id, &cleaned_html)
+        .await
+        .unwrap();
+    let markdown_id = markdown.markdown_id;
+
+    generate_and_store_summary(&state.db, markdown_id, &cleaned_html)
+        .await
+        .unwrap();
```
coreyja: All of this is what happens in the `ProcessArticle` job right?

I don't think we want to do it in the Handler, we should just do all this in the job!

The reason for that is we don't want to do all this work in the web request to add the URL, we want to do it in the 'background' which is what our job helps us do!

coreyja (changes requested):

db/src/urls.rs
Seif-Mamdouh: You correct! I thought I was suppose to call `markdown_content` as a different argument

db/src/urls.rs
Seif-Mamdouh: I thought I would have to pass down to `store_markdown_article`. I think all these silly mistakes are caused because the `db/url.rs` code is kind of all over the place. I will make sure to clean it up

cli/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 
 [dependencies]
 tokio = { workspace = true, features = ["full"] }
-db = { path = "../db" }
+cores = { path = "../cores" }
```
Seif-Mamdouh, line 10: changed `db` folder name into 'cores'

Seif-Mamdouh:
### Pair Session Changes

`cores/src`
- changed `DB` to `Cores`
- cleaned up `cores/db`
     > category.rs
     > markdown.rs
     > page_snap_shot.rs
     > page_snap_shot.rs
- deleted `persist_article` functionallity

`cores/migrations`
- altered the `markdown migrations` to add `page_snapshot_id`


`web/src/jobs.rs`
- in the `ProcessArticle`
- Created a page_snapshot variable to insert a new page snapshot into the database and return the inserted record to add to the Markdown table, with the data later used to generate a summary.

`web/users/add_url.rs`
- called the job `insert_article_handler` directly in to the `insert_article_handler` function

`web/src/pages.rs`
- replaced the existing code block with a new one that fetches the latest PageSnapShot and its associated Markdown from the database. This ensures that the most recent snapshot and its markdown content are retrieved for the given article.

web/src/jobs/process_article.rs
```diff
@@ -25,7 +26,30 @@ impl Job<AppState> for ProcessArticle {
             .await
             .into_diagnostic()?;
 
-        persist_article(db, page).await.unwrap();
+        let raw_html = download_raw_html(&page.url).await.unwrap();
+        let url_parsed = Url::parse(&page.url).unwrap();
+        let current_time = chrono::Utc::now();
+        let cleaned_html = clean_raw_html(&raw_html, &url_parsed).unwrap();
+
+        let page_snapshot = sqlx::query_as!(
```
Seif-Mamdouh, line 34: Attempt to insert a new page snapshot into the database and return the inserted record.

web/src/pages.rs
```diff
@@ -52,26 +52,35 @@ pub async fn article_detail(
...
-        article.markdown_id
+    let page_snapshot = sqlx::query_as!(
+        PageSnapShot,
+        "SELECT * FROM pagesnapshot WHERE page_id = $1 ORDER BY fetched_at DESC LIMIT 1",
+        article.page_id
     )
     .fetch_optional(&state.db)
     .await?;
 
+    let markdown = if let Some(page_snapshot) = page_snapshot {
```
Seif-Mamdouh, line 69: fetching the latest PageSnapShot and its associated Markdown from the database.

cores/migrations/20240628024728_AddSummaryToPageSnapShot.up.sql
```diff
@@ -0,0 +1,2 @@
+-- Add migration script here
+ALTER TABLE PageSnapshot ADD COLUMN summary TEXT DEFAULT '' NOT NULL;
```
coreyja: Ahh we probably want to remove this migration, since we are saving the summary to `markdown` right?

I think to do that nicely you can do `sqlx migrate revert` TWICE, since there are two migrations to revert. Then you can delete this up and the down and re-run migrate to add the summary to markdown!

deny.toml
```diff
@@ -76,14 +76,14 @@ notice = "deny"
...
 ]
-# Threshold for security vulnerabilities, any vulnerability with a CVSS score
+# Threshold for security vulnerabilities, any vulnerability with a CVSS scores
 # lower than the range specified will be ignored. Note that ignored advisories
 # will still output a note when they are encountered.
-# * None - CVSS Score 0.0
-# * Low - CVSS Score 0.1 - 3.9
-# * Medium - CVSS Score 4.0 - 6.9
-# * High - CVSS Score 7.0 - 8.9
-# * Critical - CVSS Score 9.0 - 10.0
```
coreyja: Haha not a big deal but I think these changes are all from the find and replace, so might be good to revert the changes to this file

coreyja:
This is looking good! Just that one migration thing and I think we are good to go!

deny.toml
```diff
@@ -79,11 +79,11 @@ ignore = [
 # Threshold for security vulnerabilities, any vulnerability with a CVSS scores
 # lower than the range specified will be ignored. Note that ignored advisories
 # will still output a note when they are encountered.
-# * None - CVSS Scores 0.0
-# * Low - CVSS Scores 0.1 - 3.9
-# * Medium - CVSS Scores 4.0 - 6.9
-# * High - CVSS Scores 7.0 - 8.9
-# * Critical - CVSS Scores 9.0 - 10.0
+# * None - CVSS Score 0.0
```
Seif-Mamdouh, line 82: changed it back 😂.

cores/migrations/20240628024728_AddSummaryToPageSnapShot.up.sql
```diff
@@ -1,2 +0,0 @@
--- Add migration script here
-ALTER TABLE PageSnapshot ADD COLUMN summary TEXT DEFAULT '' NOT NULL;
```
Seif-Mamdouh: removed `Summary` Column from `PageSnapshot` table and delete the `.up` && `.down`

coreyja (approved):
Awesome! Gonna merge this!

PR #19, opened 2024-06-30
Add Test Error Routes under an Admin Namespace

coreyja:
I wanted to make sure that the Sentry reporting is working like I wanted for Errors, and the best way to do this was to just create a page that always errored!

So this creates two admin routes that check a list of usernames that will be set as an env var, and if the user is an Admin it fires some test errors
If the user is not an admin we return a 401 UNAUTHORIZED http response

### DIFF
--- a/web/src/admin.rs
+++ b/web/src/admin.rs
@@ -1,19 +1,76 @@
-use axum::{extract::State, routing::get};
+use axum::{
+    extract::{FromRequestParts, State},
+    http::{request::Parts, StatusCode},
+    response::IntoResponse,
+    routing::get,
+};
 use cja::app_state::AppState as _;
 use color_eyre::eyre::Context;
+use db::users::User;
 
-use crate::{AppState, WebResult};
+use crate::{err, users::ExtractUserError, AppState, WebResult};
 
-async fn simple_error() -> WebResult<()> {
+async fn simple_error(_: AdminUser) -> WebResult<()> {
     Err(color_eyre::eyre::eyre!("This is a test error"))?
 }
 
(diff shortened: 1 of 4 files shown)


web/src/admin.rs
```diff
@@ -0,0 +1,91 @@
...
+    Err(color_eyre::eyre::eyre!("This is a test error"))?
+}
+
+async fn sql_error(_: AdminUser, State(app_state): State<AppState>) -> WebResult<()> {
+    let _ = sqlx::query("This is not valid sql")
+        .fetch_one(app_state.db())
+        .await
+        .wrap_err("We meant for this to fail but lets see the error chain")?;
+    Ok(())
+}
```
coreyja, line 23: These are two routes that are error on purpose to test error reporting and things like that!

I want to see these show up in Sentry.io

web/src/routes.rs
```diff
@@ -2,13 +2,15 @@ use axum::routing::get;
...
 use crate::{
+    admin,
     pages::{home, landing, user_dashboard},
     users, AppState,
 };
 
 pub fn routes(app_state: AppState) -> axum::Router {
     axum::Router::new()
         .route("/", get(home))
+        .nest("/_", admin::routes())
```
coreyja, line 13: This allows us to define all the admin routes in the admin module just to organize things a bit!

We might break down more things like this later!

coreyja:
Tagging you here @Seif-Mamdouh cause I want you to at least be able to see this but I'm going to merge this early so it deploys cause I want to test it in prod lol

PR #20, opened 2024-07-01
End2End Tests with Playwright

coreyja:
This adds End2End Browser tests with Playwright!

These run in multiple browsers and perform actions like a User to make sure the app is working!

Right now we are testing sign up and login, and was some tests of summarization marked as skipped cause they are having issues.

BUT this PR also improves the error handling a bit and uses color_eyre everywhere, and I want to lean into that!


scripts/setup.sh
```diff
@@ -6,6 +6,7 @@ pushd $(git rev-parse --show-toplevel) > /dev/null
 
 cargo install sqlx-cli --version 0.7.4  --no-default-features --features postgres,rustls
 cargo sqlx database create
+DATABASE_URL="${DATABASE_URL}_test" cargo sqlx database create
```
coreyja, line 9: Oh we don't need this! We do it before we run tests in Playwright now, will remove

web/src/sessions.rs
```diff
@@ -47,7 +47,7 @@ impl Session {
             cja::tower_cookies::Cookie::build(("session_id", session.session_id.to_string()))
                 .path("/")
                 .http_only(true)
-                .secure(true)
+                .secure(std::env::var("HTTPS").unwrap_or_else(|_| "true".to_string()) == "true")
```
coreyja, line 50: Safari doesn't allow secure cookies on localhost. Most other browsers treat localhost as a 'secure' set and allow you to use secure cookies like you could on a `HTTPS` site.

Safari wasn't in the testing, so this creates a way for testing to NOT set a secure cookie.

This is kinda annoying... But its easier than getting CI to run with a nice https certificate right now!

PR #22, opened 2024-07-09
Ticket 296 view articles by category

Seif-Mamdouh:
In this PR, we provided a feature that allows the user to browse a list of their categories and click on a specific category to get its summary.

### Key Changes
- `cores/migrations/`:
    - We created `CategoryMarkdown` table to establish a many-to-many relationship between 
    categories and markdown content.
    - The reason why I chose this approach is because a `markdown_id.summary` can have 
      more than `category` and vice versa.
    - For example a category `'Data Flow'` can have multiple `summaries`
    - added a `category_id` in `category` table

- `cores/src/markdown.rs`:
   - added a `query` to insert a record into the `CategoryMarkdown` table

- `web/src/pages.rs`
   -  added a handler `my_categories` that list all categories
   -  added another handler `articles_by_category` that shows the summary related to the 
      category

- `web/src/routes.rs`
     - added two `routes` 
     - 1 for `user_categories`: listing all categories associated for the user 
     - 2 for `articles_by_category`: listing all the summaries for a certain category

### Demo
https://github.com/coreyja/knowledge/assets/55955558/a1a1c424-b3e8-444a-a67d-6b39ed03849a

### `MarkdownCategory Table` Rundown
Screenshot 2024-07-09 at 9 10 32 AM


cores/src/markdown.rs
```diff
@@ -35,7 +42,17 @@ pub async fn store_in_markdown_table(
     let category = generate_categories(&content_md).await?;
     let category_result =
         store_in_category_table(pool, markdown_result.markdown_id, &category).await?;
-    println!("Category result: {category_result:?}");
+
+    let category_markdown_result = sqlx::query_as!(
```
Seif-Mamdouh, line 46: Added the query to insert the `markdown_id` && `category_id` to create many to many relationships!

web/src/pages.rs
```diff
@@ -146,3 +151,68 @@ pub async fn my_articles(
     })
     .into_response())
 }
+
+#[axum::debug_handler(state = AppState)]
+pub async fn my_categories(
```
Seif-Mamdouh, line 156: handler to get `categories`

web/src/pages.rs
```diff
@@ -146,3 +151,68 @@ pub async fn my_articles(
...
+#[axum::debug_handler(state = AppState)]
+pub async fn articles_by_category(
+    t: Template,
+    Path(category_id): Path<Uuid>,
+    State(state): State<AppState>,
+    user: User,
+) -> WebResult<Response> {
+    info!("Fetching articles for category_id: {}", category_id);
+
+    let articles = sqlx::query!(
```
Seif-Mamdouh, line 199: In this query we connect related data across multiple tables `category`, `markdown`, `pagesnapshot`, and `pages` to retrieve all page information.

We verify that a certain `summary` belong to a certain `category`

web/src/pages.rs
```diff
@@ -160,9 +160,18 @@ pub async fn my_categories(
 ) -> WebResult<Response> {
     info!("Fetching categories for user_id: {}", user.user_id);
 
-    let categories = sqlx::query!("SELECT * FROM category",)
-        .fetch_all(&state.db)
-        .await?;
+    let categories = sqlx::query!(
```
Seif-Mamdouh, line 163: Noticed I was selecting all categories regardless of the `user_id`. So I changed it to make sure to selecting all `category_id` with the designated `user_id`

web/src/pages.rs
```diff
@@ -146,3 +151,77 @@ pub async fn my_articles(
...
+    )
+    .fetch_all(&state.db)
+    .await?;
+
+    Ok(t.render(maud::html! {
+        h1 { "Articles in Category" }
+        ul {
+            @for article in articles {
+                li {
+                    a href=(article.url) { (article.url) }
```
coreyja: Can we link to the `articles/:id` here instead of directly to the actual article URL?

That way you can click on the link to get our 'detail' pages!

cores/migrations/20240708182646_AddCategoryMarkdown.up.sql
```diff
@@ -0,0 +1,15 @@
+-- Drop the existing primary key constraint
+ALTER TABLE Category DROP CONSTRAINT IF EXISTS category_pkey;
+
+-- Ensure Category table has category_id column with a default UUID value
+ALTER TABLE Category ADD COLUMN IF NOT EXISTS category_id UUID DEFAULT gen_random_uuid() NOT NULL;
+
+-- Set category_id as the new primary key
+ALTER TABLE Category ADD CONSTRAINT category_pkey PRIMARY KEY (category_id);
```
coreyja: I'm a _bit_ worried about this... 🤔 

Do you know/expect the Category table in Prod to be empty? If so then this is probably fine! But if not I don't think we can change out the Primary Key like this 🤔

cores/migrations/20240708182646_AddCategoryMarkdown.up.sql
```diff
@@ -0,0 +1,15 @@
+-- Drop the existing primary key constraint
+ALTER TABLE Category DROP CONSTRAINT IF EXISTS category_pkey;
```
coreyja: And this column shouldn't exist in the Category table before this migration right? Not awful to leave it in, but found it a touch confusing since I didn't expect this column to exist when this migration runs lol

cores/migrations/20240708182646_AddCategoryMarkdown.down.sql
```diff
@@ -0,0 +1,2 @@
+-- Add migration script here
+DROP TABLE IF EXISTS CategoryMarkdown;
```
coreyja, line 1: Wonder if we also need/want to drop the `Category.category_id` column in this `down`, since it got added in the `up` right?

cores/migrations/20240708182646_AddCategoryMarkdown.up.sql
```diff
@@ -0,0 +1,15 @@
...
+
+-- Ensure Category table has category_id column with a default UUID value
+ALTER TABLE Category ADD COLUMN IF NOT EXISTS category_id UUID DEFAULT gen_random_uuid() NOT NULL;
```
coreyja: Oh nvmd about the table being empty! This will fill in any rows without an ID

Just so you know, this is _technically_ not a 'safe' operation and could result in a bit of downtime for the site.
Since to add this column and add the default Postgres has to lock this whole table so nobody else can use it

In this case thats gonna be fine, prod is basically not used so it can handle 5 seconds of "downtime" lol But good to know for the future!

Eventually we will want to add this as nilable and backfill it in code, before we go back and mark the column as non-nil

But thats a lot of work and I think this is fine for now 😆

web/src/pages.rs
```diff
@@ -50,6 +50,10 @@ pub async fn user_dashboard(t: Template, user: User) -> TemplatedPage {
             a href="/articles" { "My Articles" }
         }
 
+        h3 {
+            a href="/categories" { "My Categories" }
```
coreyja, line 54: 💯

web/src/pages.rs
coreyja: Love this! Nice job on all the joins lol

web/src/pages.rs
```diff
@@ -146,3 +151,77 @@ pub async fn my_articles(
...
+    t: Template,
+    State(state): State<AppState>,
+    user: User,
+) -> WebResult<Response> {
+    info!("Fetching categories for user_id: {}", user.user_id);
+
+    let categories = sqlx::query!(
+        "SELECT DISTINCT c.* 
+     FROM category c
+     JOIN categorymarkdown cm ON c.category_id = cm.category_id
```
coreyja: I think we can simplify the join syntax just a touch here! And this is one reason we wanted the column names to 'match' on both tables!

We should be able to do something like this instead of doing the full `ON` stuff!

```suggestion
     JOIN categorymarkdown cm USING (category_id)
```

But its just syntactic sugar (just looks prettier lol) does the same thing under the hood!

web/src/pages.rs
```diff
@@ -146,3 +151,77 @@ pub async fn my_articles(
...
+        user.user_id
+    )
+    .fetch_all(&state.db)
+    .await?;
+
+    Ok(t.render(maud::html! {
+        h1 { "My Categories" }
+        ul {
+            @for category in categories {
+                @let category_name = category.category.unwrap_or("No category".to_string());
```
coreyja, line 180: Hmm why is `category.category` an Option?

If we created a `Category` row in the table wouldn't we expect it to have a Category? Is that column null-able and if so should it be lol?

coreyja (changes requested):
This looks GREAT!!

Love the routing and linking, that all looks perfect!
And your SQL joins look great too, really well done!

Most of my comments are small potatoes! The meat of this is awesome!

web/src/pages.rs
Seif-Mamdouh: because we have column name in `category` table  that is named `category`. I'm trying to access the category column. 

Is there a cleaner way to write it?

Screenshot 2024-07-09 at 11 08 39 AM

cores/migrations/20240708182646_AddCategoryMarkdown.up.sql
Seif-Mamdouh: I updated the migration to add new category_id column without setting it to the PK, this is a better approach so we have a smooth data flow.

cores/migrations/20240708182646_AddCategoryMarkdown.up.sql
```diff
@@ -1,13 +1,16 @@
--- Drop the existing primary key constraint
-ALTER TABLE Category DROP CONSTRAINT IF EXISTS category_pkey;
+-- Add category_id column ONLY if it doesn't exist
```
Seif-Mamdouh, line 1: I think this appraoch is safer because we don't alter the primary key to main the current data relationship.

I added a `catgory_id` column as a unique identifier

Let me what you think about this appraoch 😆

coreyja: This is PERFECT!

Love each step that you took here!

In a "big paranoid" company you might see teams split this into multiple migrations/deploys so they can watch each step. But I think this is great, more just chatting!

web/src/pages.rs
coreyja: Ahh the answer to my question was that the column is marked as `is_nullable`, and thats why we get an option back!

We might eventually want to make that non-nullable, cause if there is a category row, I think there should be something in the category column lol
But that doesn't have to be a thing for this PR!

coreyja (approved):
Love this! Merging it!

PR #23, opened 2024-07-09
added functionality to parse titles and diply on the web UI

Seif-Mamdouh:
### New feature capabilities 
This PR adds functionality to store the article title in the `markdown` table and display the `article title` when the user views articles through their `My Articles` or `My Categories` tabs.

### Key File Changes

`cores/src/page_snapshot.rs`
- added function `extract_title` that parse the html content and look for the `title` HTML tag
- Utlized the `Scraper` built in functionality `Selector` to parse the title from the html 
- docs ref: https://docs.rs/scraper/0.19.0/scraper/index.html#parsing-a-selector

`cores/src/markdown.rs`
- updated the query in the to `store_in_markdown_table` function to add a `title` to the `DB`

`web/src/pages.rs`
- added functionality to display the title in both `My Articles` && `My Categories` tabs.


cores/src/markdown.rs
```diff
@@ -27,14 +27,17 @@ pub async fn store_in_markdown_table(
 ) -> color_eyre::Result<Markdown> {
     let markdown_id = Uuid::new_v4();
     let content_md = html2md::parse_html(&page_snapshot.cleaned_html);
+    let title = extract_title(&page_snapshot.raw_html);
+    println!("Title: {title:?}");
```
Seif-Mamdouh, line 31: parsing the title from `raw_html`

cores/src/page_snapshot.rs
```diff
@@ -24,3 +26,14 @@ pub fn clean_raw_html(raw_html: &str, url: &Url) -> color_eyre::Result<String> {
...
 }
+
+pub fn extract_title(content: &str) -> Result<String> {
+    let document = Html::parse_document(content);
+    let title_selector = Selector::parse("title").unwrap();
+
+    if let Some(title_element) = document.select(&title_selector).next() {
+        Ok(title_element.text().collect::<String>())
+    } else {
+        Ok("No title found".to_string())
```
coreyja: This could probably return an `Result<Option<String>>` so that in this case someone who calls this function and react how they want to a title not being found without having to look for this specific error string!

coreyja (commented):
Nice!!

This was a bit easier than I was thinking it might be 😆 

Only comment is about returning an `Option<String>` for the title since there might not be one in the HTML

coreyja (approved):
🎉

PR #24, opened 2024-07-25
Refactor DB Models and improve Details Page a Bit

coreyja:
This cleans up the DB a bit. Renames some tables to be the plural version, and fixes a few of the relations. Also changes the null-ability of a few columns to make how we are using them.

Will leave some comments inline!


cores/migrations/20240725012639_MakeCategoryCategoryNonNullable.up.sql
```diff
@@ -0,0 +1,5 @@
+-- Add migration script here
+ALTER TABLE category
+ALTER COLUMN category
+SET
+  NOT NULL;
```
coreyja, line 5: No reason to have a category row without a category string, so set it to not null

cores/migrations/20240725012747_RenameTablesToPlural.up.sql
```diff
@@ -0,0 +1,13 @@
...
+RENAME TO categories;
+
+ALTER TABLE markdown
+RENAME TO markdowns;
+
+ALTER TABLE pagesnapshot
+RENAME TO page_snapshots;
+
+ALTER TABLE categorymarkdown
+RENAME TO markdown_categories;
```
coreyja, line 13: Make all the tables plural

And for some reason I like markdown_categories over category_markdowns lol 🤷  just sounds nicer to me

cores/migrations/20240725014146_FixCategoryRelations.up.sql
```diff
@@ -0,0 +1,3 @@
+-- Add migration script here
+ALTER TABLE categories
+DROP COLUMN markdown_id;
```
coreyja, line 3: We have the markdown_categories join table that associates a markdown with multiple categories so we don't want this direction FK too

cores/migrations/20240725015052_AddCategoryPrimaryKey.up.sql
```diff
@@ -0,0 +1,5 @@
+-- Add Primary Key to Categories on category_id
+ALTER TABLE categories ADD PRIMARY KEY (category_id);
+
+ALTER TABLE categories
+DROP CONSTRAINT category_id_unique CASCADE;
```
coreyja, line 5: This had a unique index but it wasn't a Primary Key, so we add a primary key and then drop the duplicate unique index/constraint

cores/migrations/20240725020036_MakeSummaryNullable.up.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE markdowns
+AlTER COLUMN summary
+DROP NOT NULL;
+
+ALTER TABLE markdowns
+AlTER COLUMN summary
+DROP DEFAULT;
```
coreyja, line 7: Summary was a non-null column with an empty string default

But we make a markdowns before we have a summary, so it make more sense for it to be nullable and not have a default empty string value

cores/migrations/20240725022254_MakeFetchedAtNonNull.up.sql
```diff
@@ -0,0 +1,5 @@
+-- Add migration script here
+ALTER TABLE page_snapshots
+ALTER COLUMN fetched_at
+SET
+  NOT NULL;
```
coreyja, line 5: We save a page_snapshots when we have fetched the html so there isn't a chance for this to be null

web/src/jobs/process_article.rs
```diff
@@ -43,11 +43,8 @@ impl Job<AppState> for ProcessArticle {
         .await?;
 
         let markdown = store_in_markdown_table(db, page_snapshot).await?;
-        let markdown_id = markdown.markdown_id;
 
-        generate_and_store_summary(db, markdown_id, &cleaned_html).await?;
-
-        generate_categories(&markdown.summary).await.unwrap();
```
coreyja, line 50: We weren't saving these results anywhere so no need to make this call

web/src/pages.rs
```diff
@@ -90,35 +90,50 @@ pub async fn article_detail(
...
         }
     };
 
-    Ok(t.render(rendered_html).into_response())
+    Ok(t.render(maud::html! {
+        h1 { "Article Detail" }
+        h2 { "URL: " (article.url) }
+        div { (inner_html) }
+    })
+    .into_response())
```
coreyja, line 136: I don't really like the structure how this method...

Just the nested `if lets` and stuff makes it hard to follow

Just saying that I want to come back to how I structure these methods as I made more of them

web/src/pages.rs
```diff
@@ -162,11 +177,11 @@ pub async fn my_categories(
...
-         FROM category c
-         JOIN categorymarkdown cm ON c.category_id = cm.category_id
-         JOIN markdown m ON cm.markdown_id = m.markdown_id
-         JOIN pagesnapshot ps ON m.page_snapshot_id = ps.page_snapshot_id
-         JOIN pages p ON ps.page_id = p.page_id
+         FROM categories c
+         JOIN markdown_categories cm USING (category_id)
+         JOIN markdowns m USING (markdown_id)
+         JOIN page_snapshots ps USING (page_snapshot_id)
+         JOIN pages p USING (page_id)
```
coreyja, line 184: I think this `USING` syntax is nice and we specifically named the columns so that we could use it!

web/src/pages.rs
```diff
@@ -218,9 +233,13 @@ pub async fn articles_by_category(
             @for article in articles {
                 li {
                     a href=(format!("/articles/{}", article.page_id)) { (article.url) }
-                    p { b { "Title: " } (article.title.as_deref().unwrap_or("No title")) }
-                    p { b { "Category: " } (article.category.as_deref().unwrap_or("No category")) }
-                    p { b { "Summary: " } (article.summary) }
+                    @if let Some(title) = article.title {
```
coreyja, line 236: Instead of using `unwrap_or` just don't render the element if its missing

cores/src/category.rs
```diff
@@ -6,22 +6,19 @@ use crate::openai_utils::generate_embedding;
...
 pub async fn store_in_category_table(
     pool: &PgPool,
-    markdown_id: Uuid,
     category: &str,
 ) -> color_eyre::Result<Category> {
     let embedding = generate_embedding(category).await?;
     let result = sqlx::query_as!(
         Category,
-        "INSERT INTO Category (category_id, markdown_id, category, embedding) VALUES ($1, $2, $3, $4) RETURNING *",
+        "INSERT INTO categories (category_id, category, embedding) VALUES ($1, $2, $3) RETURNING *",
```
Seif-Mamdouh, line 20: Do you think we don't need to have `markdown_id` anymore? The reason why we needed the `markdown_id` is because we wanted a `FK` reference to generate `Categories` to a certain `MD_id`

web/src/jobs/process_article.rs
Seif-Mamdouh: I was calling `generate_categories` so it generates categories as soon the the user inserts a URL onto the Web app. 

And then we were storing the category in the `Category Table`

Then later we rendering it on the HTML. We don't need to do that anymore 🤔

cores/src/category.rs
coreyja: We don't need _this_ FK cause we already have another one 🎉 

There is the `markdown_categories` table that has two columns `category_id` and `markdown_id`

And we use this to create the links! This way a Markdown can link to multiple categories in the DB! [We don't do this yet but I think we will soon!]

web/src/jobs/process_article.rs
coreyja: This call I removed I think is just 'duplicative'!

We do the Category stuff in `store_in_markdown_table` and we actually save it to the DB there

cores/src/category.rs
Seif-Mamdouh: Oh okay, so it's like a `JOIN` Table?

Seif-Mamdouh (approved):

